PR #33, opened 2023-03-24
added default theme setting

ayushanand308:
Only did a very minor change to set the default theme of the app to system's default theme.

### DIFF
--- a/app/src/main/kotlin/org/scribe/activities/MainActivity.kt
+++ b/app/src/main/kotlin/org/scribe/activities/MainActivity.kt
@@ -23,6 +23,9 @@ class MainActivity : SimpleActivity() {
         scribe_key.setOnClickListener {
             (getSystemService(INPUT_METHOD_SERVICE) as InputMethodManager).showInputMethodPicker()
         }
+
+
+
     }
 
     override fun onResume() {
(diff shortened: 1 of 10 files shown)


ayushanand308:
Actually only the functionality of changing theme according to user's default theme was added as you asked for that feature as well. I was not able to figure out a way to make the theme change when the dark theme check box is checked and I am still trying for that. Any help or feedback will be appreciated.

andrewtavis:
Hmmm, interesting 🤔 Even when I change my default theme it's not changing though 🤔 I saw in your video that it is working, but then it doesn't seem to be working on my end. What device/emulator are you using?

ayushanand308:
@andrewtavis I am using Xiaomi Redmi 9.

andrewtavis:
I’ll try on that one and see if the change goes though. Thanks, @ayushanand308!

andrewtavis (approved):
@ayushanand308, want to apologize for my lack of responsiveness on this. Is a very busy time for me right now between a new job and balancing other projects. Those are excuses though, and I should have at the very least been able to figure out that I don't have the capacity for this project. Really do want Scribe-Android to exist, but it's still going to be a while before we can adequately put time into this.

With the above being said, these changes do appear to be working on the Xiaomi that you tested them on, so I'm fine with merging. The issue will need to stay open as I can't see the changes on Pixels among other emulators, but that's no reason for this contribution to not come in.

Again my apologies. Wishing you the best, and maybe we can find another time to really look into all of this :)

Issue #34, opened 2023-07-03
Look into AOSP as a basis


andrewtavis:
### Terms

- [X] I have searched [open and closed issues](https://github.com/scribe-org/Scribe-Android/issues?q=is%3Aissue)
- [X] I agree to follow Scribe-Android's [Code of Conduct](https://github.com/scribe-org/Scribe-Android/blob/main/.github/CODE_OF_CONDUCT.md)

### Issue

Something that was just discussed is using [AOSP’s Android keyboard](https://www.apkmirror.com/apk/lineageos/android-keyboard-aosp-2/) as a basis for Scribe-Android. Looking into this would be something we should do 😊

andrewtavis:
I’d say that maintaining the current code base makes some sense given the app structure and such, but maybe AOSP has a convenient way of making an installable Android keyboard.

angrezichatterbox:
Could I do any help in this issue .

andrewtavis:
This would be an interesting thing to check out, @angrezichatterbox! Maybe you could research it a bit and we could close this it it's not possible. We can maybe do a call in a few weeks to discuss what's next? Happy to plan things out for you!

And I'll get to the current PRs eventually. Really appreciate your consistency here. After this weekend I should have a bit more time :)

andrewtavis:
Assigned, and let me know what you find here!

angrezichatterbox:
I am not able to install the aosp keyboard apk on my android simulator .. However I looked onto some of it features and also I checked out on the fork projects of aosp keyboard called simple keyboard which was able to install and got some similarities  between the current scribe keyboard features and simple keyboard However when I searched about aosp keyboard I got to know that it has gesture control and also word correction . However I wasn't able to test any of these features since simple keyboard didn't have it . I would continue to try to install the aosp keyboard

andrewtavis:
Hey @angrezichatterbox 👋 Scribe-Android is based off of Simple Keyboard, so the functionality that that one has should also generally be over here. I don't think there were too many changes from Simple Keyboard from when I cloned it, and I know that that project now has license issues, but it's also a thought to see what else we could bring over from that project from before the license switched. Let's be in touch on this!

andrewtavis:
Thanks so much for checking!

angrezichatterbox:
Simple keyboard was quite similar to Scribe even now .  I would check if I could get aosp apk on my android simulator to check the various features it offers

andrewtavis:
[...]

PR #35, opened 2024-06-03
feat:Added a shadow beneath the key

angrezichatterbox:


andrewtavis:
Finally able to test this a bit, @angrezichatterbox :) So generally it's working, but the space bar being made smaller by padding isn't ideal. Ideally we could get these keys looking like the iOS ones with rounding and a bit of space between the keys. Do you think that that would be possible? We can also hold off on the edits until you're done with the research in #34 :)

angrezichatterbox:
Ok I will fix it after I look into the AOSP keyboard and will also look into the other issues ..

angrezichatterbox:
I am free to discuss any changes that is required to be made in this project .

andrewtavis:
Sounds great :) Let's check in maybe once you have some thoughts on #34!

angrezichatterbox:
Sorry for the delay in checking and fixing the issues as well the AOSP thing I had exams I will return to correcting those from tmrw

andrewtavis:
100% ok, @angrezichatterbox! Good luck with exams :)

angrezichatterbox:
I have made some changes to the code and made the keys more rounded also I fixed the issue of space bar having rounded white curbed rectangle in between.

andrewtavis (approved):
Hey @angrezichatterbox 👋 All's looking really good here! I'll bring this in and write in the issue about the next steps :)

PR #36, opened 2024-06-16
Period on Double Tap on Space Bar

angrezichatterbox:
This PR adds the feature to the keyboard where the user on double tap on the space bar would be able to see a period similar to other keyboards


andrewtavis (approved):
Hey @angrezichatterbox 👋 Post needing to do a full Android Studio re-install - my settings were out of date - I can confirm that this is exactly what we're looking for :) Thanks so much for working on this and for your patience!

Issue #27, opened 2022-05-19
Add dark and light mode responsiveness


andrewtavis:
### Terms

- [x] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Android/issues?q=is%3Aissue+label%3Afeature)
- [x] I agree to follow Scribe-Android's [Code of Conduct](https://github.com/scribe-org/Scribe-Android/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue is to implement color color scheme switching between light and dark modes.

- This issue is blocked by #25 as the schemes should be created before this issue

ayushanand308:
@andrewtavis This seems a good feature. Will be working on it :).

andrewtavis:
Hey @ayushanand308! Thanks for your interest in Scribe! Yes I think this would be a great feature to better understand the project 😊

I'll assign this to you, and please let me know if you have any questions!

ayushanand308:
@andrewtavis sure!

ayushanand308:
@andrewtavis Basically what I understand from the task description is that I have to implement a way to switch between dark and light themes on the keyboard right?

ayushanand308:
![WhatsApp Image 2023-03-01 at 00 24 47](https://user-images.githubusercontent.com/99035102/221951891-e5b5a478-3e67-4cb2-b674-5ff87ed5f46d.jpg)



Thinking of implementing it like this. Is this ok?

andrewtavis:
Yes, exactly @ayushanand308! Specifically it would also be good if the default was set based on the mode of the user’s phone. Not sure if you have an idea on how best to do that, and I could support with exploration if needed :) So if the user has dark mode on their phone, it would be great if the option was already selected for them 😊

ayushanand308:
@andrewtavis I would surely ask for help wherever I need it.

andrewtavis:
Wonderful, @ayushanand308 :) Looking forward to the results! 😊

ayushanand308:
@andrewtavis 
I am not able to understand how to add functionality to this dark theme checkbox toggle like there is for "vibrate on keypress" and others. Can you help with this.:-)

https://user-images.githubusercontent.com/99035102/222565267-8374282c-f3d2-45a1-8279-1c961c63bb74.mp4

andrewtavis:
Hi @ayushanand308! Just wanted to write that I have this in mind. Currently visiting family at the moment :)

[...]

PR #38, opened 2024-07-04
feat:Added Dark Mode for the app as well as the keyboard

angrezichatterbox:
[x] This pull request is on a [separate branch](https://docs.github.com/en/get-started/quickstart/github-flow) and not the main branch

---

### Description
The Pull request aims to add dark mode and light mode functionality to the application.The current approach adds a dark mode or light mode on the user clicking the checkbox which is present in the app settings. 

The changes have been tested the keyboard changes it color to a darker color when the user check the dark mode option in the settings . the application also changes to the dark mode theme given in the scheme present for the dark and light mode respectively 
### Related issue
-   #27

### DIFF
--- a/app/src/main/kotlin/org/scribe/activities/MainActivity.kt
+++ b/app/src/main/kotlin/org/scribe/activities/MainActivity.kt
@@ -1,6 +1,8 @@
 package org.scribe.activities
 
+import android.content.Context
 import android.content.Intent
+import android.content.res.Configuration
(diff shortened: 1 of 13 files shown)


andrewtavis (approved):
Working really well, @angrezichatterbox! There are now a few things in the app that even the iOS app doesn't have 😊 In iOS the double space period shortcut is always on, and the app is reactive, but they can't change the color mode themselves. Will write a bit more in the issue about next steps!

Issue #42, opened 2024-07-27
Switch app identifier to be.scri from org.scribe


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Android/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-Android's [Code of Conduct](https://github.com/scribe-org/Scribe-Android/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue is a simple but very important one. We now have the URL scri.be, so we should be using this in the identifiers of our applications. The Scribe-Android identifier should be switched from `org.scribe` to `be.scri` to be in line with our web presence. Maybe this change will be reflected on Scribe-iOS at some point too :)

### Contribution

Happy to review the changes here!

andrewtavis:
Ping @angrezichatterbox 👋 Do you want to get this done, and then we can move on to others? :)

angrezichatterbox:
Ya sure I will do it.

andrewtavis:
Thanks, @angrezichatterbox! :)

angrezichatterbox:
I have created a pull request solving this issue.

andrewtavis:
Closed via #49 🚀 Thanks for this, @angrezichatterbox! Let me know if any of the new issues sound appealing. I think that once we have #48 done you could then make some issues to add individual menu elements?

PR #49, opened 2024-07-27
Change Application identifier

angrezichatterbox:
### Description

The pull request changes the application identifier from org.scribe to be.scri. 

### Related issue
-   #42 

- #### Additional information
1. For some reason the old method of checking of whether it was a new user and telling the user to go to settings didn't seem to work so I had to make the check using the package name of those input picker additions. Should I revert that change. ?


andrewtavis:
Hey @angrezichatterbox 👋

> For some reason the old method of checking of whether it was a new user and telling the user to go to settings didn't seem to work so I had to make the check using the package name of those input picker additions. Should I revert that change. ?

We actually discussed in the sync today, and we would prefer for the app to not force the user to install a keyboard. We'll direct them to install the keyboard on the installation page, and keyboard settings will be conditional on them being installed. It should be ok to revert that change so that the app opens without the tooltip telling the user to install the keyboard. Can you remove the code for that?

angrezichatterbox:
> Hey @angrezichatterbox 👋
> 
> > For some reason the old method of checking of whether it was a new user and telling the user to go to settings didn't seem to work so I had to make the check using the package name of those input picker additions. Should I revert that change. ?
> 
> We actually discussed in the sync today, and we would prefer for the app to not force the user to install a keyboard. We'll direct them to install the keyboard on the installation page, and keyboard settings will be conditional on them being installed. It should be ok to revert that change so that the app opens without the tooltip telling the user to install the keyboard. Can you remove the code for that?

I have reverted the change.

andrewtavis:
Checking it right now :) On build I'm getting the following error:

> Activity class {org.scribe.debug/be.scri.activities.SplashActivity.Orange} does not exist

Maybe there's a spot where it's still `org.scribe` in the code?

angrezichatterbox:
> Checking it right now :) On build I'm getting the following error:
> 
> > Activity class {org.scribe.debug/be.scri.activities.SplashActivity.Orange} does not exist
> 
> Maybe there's a spot where it's still `org.scribe` in the code?

I have replaced it now in the latest commit to wherever I found org.scribe. I'm sorry that I didn't check again. the refractor of Android Studio does not change it everywhere. Could you try to invalidate the cache and rebuild it as  I am not able to replicate the same error.

andrewtavis:
Sure, will do a little later today, @angrezichatterbox!

andrewtavis (approved):
I'm still getting the requirement to install a keyboard at the start, but it's all good for now.  We can make an issue for that specifically :)

Invalidating the cache did get rid of the error :) Thanks for this, @angrezichatterbox! 😊

Issue #48, opened 2024-07-27
Plan out and implement customizable application table views


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Android/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-Android's [Code of Conduct](https://github.com/scribe-org/Scribe-Android/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

One thing that's needed to improve the menu is that we create table views that allow for customization based on passed arguments. This issue will be a larger one, but will allow for many small issues to be made where individual menu components are added in :) Looking at the About tab from the [Figma designs](https://www.figma.com/design/c8945w2iyoPYVhsqW7vRn6/scribe_public_designs?node-id=513-2206&t=aDYg9wH7ZwFKbtAN-1) as shown below:

![About Scribe - Light](https://github.com/user-attachments/assets/0e5785de-23fb-4b17-94e0-9cfb4f74808d)

We'd need customizable cells that can have a custom icon to the left, an icon to the left that's either a right chevron, an external link icon or a counterclockwise circle, and we should also consider that we need text to the left of the right icons in some instances for menu options.

### Contribution

Happy to discuss what's needed and help implement/review! 😊

angrezichatterbox:
I implement those customizable cells where the icon is on the left and an icon on right with text in middle . Right ? .

andrewtavis:
Yes exactly, @angrezichatterbox :) It would be great if we can easily create table view cells like the ones we already have that the needed icons. It makes sense to have a common cell with options, correct? Thank you!

angrezichatterbox:
Ya it does make sense.

angrezichatterbox:
<img width="377" alt="Screenshot 2024-07-31 at 4 13 31 PM" src="https://github.com/user-attachments/assets/8a9cd94b-75fb-4e04-abfc-8f0ad3893ed0">

I have tried to my the reusable component in my personal app for testing. This is what I have so far. I have created a model that takes in 3 variables. Two images and a third button image. The button could be individually added to navigate to whatever it is prescribed to do. 
Is this what had to be done or I have to do something else as well.

angrezichatterbox:
> Ideally the entire field should be a button, @angrezichatterbox, and then the third option is an icon like the first :) Would that work?

Ya that would work and it is much easier to implement as well.

[...]

PR #50, opened 2024-08-03
feat:Implemented new about page

angrezichatterbox:

---

### Description
The pull request aims to make the About page of the Scribe Android similar to the About page of the Scribe iOS app . It also implements an adapter and a model for a recycle view that can be used to implement another such table view in other activities.  

### Related issue

- #48


andrewtavis (approved):
All's looking amazing, @angrezichatterbox! Minor things need to be changed, but we can do individual issues for those eventually. Let me know on Matrix what you're thinking about for some issues you could make 😊

Issue #44, opened 2024-07-27
Make all application strings be referenced from a Scribe-i18n XML file


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Android/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-Android's [Code of Conduct](https://github.com/scribe-org/Scribe-Android/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

Prior to release with multiple localizations, and to make it easier for people to use the exact strings that are needed for the application, it would be good to have them all be referenced from a common XML file. This file would be generated in the [Scribe-i18n](https://github.com/scribe-org/Scribe-i18n) project whenever there's a PR from Weblate and then transferred over to Scribe-Android :)

- Note that this issue is blocked by https://github.com/scribe-org/Scribe-i18n/issues/27

### Contribution

Would be happy to discuss and support this as well as review! 😊

andrewtavis:
@angrezichatterbox, would this one make sense now? You could do a PR for some Python for https://github.com/scribe-org/Scribe-i18n/issues/27 where you'd make the XML equivalent of the [xcstrings to json](https://github.com/scribe-org/Scribe-i18n/tree/main/Scribe-i18n/Scripts) scripts, and then from there we can start using the XML files and don't need to worry about putting them all in later on?

angrezichatterbox:
Yes, I would work on this issue. I would work on writing Python scripts for JSON to XML and XML to JSON.

angrezichatterbox:
I don't have much idea about how localization is in iOS. But I haven't seen a single strings.xml file for all the languages in Android yet. So could I put each language into its own XML or should I look into having a single strings.xml file for all languages?

andrewtavis:
Will quote you over there, @angrezichatterbox :)

andrewtavis:
Ok, @angrezichatterbox, Scribe-i18n is now in the repo at [main/app/src/main/res/i18n](https://github.com/scribe-org/Scribe-Android/tree/main/app/src/main/res/i18n). Let me know if accessing the values there can work out, and we can also change up any of the naming of files if you think that would help :)

angrezichatterbox:
[...]

PR #57, opened 2024-08-14
Reference strings from i18n folder 

angrezichatterbox:
### Contributor checklist

---

### Description

The strings will now be referenced from the i18n folder, which is a subtree of Scribe-i18n. The About page is fully localized, as it was recently created and all the strings were available. The bottom navigation bar is also localized. However, the other pages could not be localized because the strings were not present. As a result, the old strings.xml is still present. I will create a document for the strings currently referenced in the various files from strings.xml.

### Related issue

-   #44

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -12,6 +12,8 @@ if (keystorePropertiesFile.exists()) {
 android {
     compileSdk 34
 
+
+
     defaultConfig {
         applicationId "be.scri"
         minSdk 23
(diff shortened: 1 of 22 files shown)


andrewtavis:
Thanks for this, @angrezichatterbox! I'll try to get the review done this evening :)

andrewtavis (approved):
Really amazing to see Scribe-Android adopt things that are so new to Scribe-iOS so early in its life, @angrezichatterbox 😊 Thanks so much for this! So great that Scribe-i18n can already fulfill it's purpose :) :)

Issue #46, opened 2024-07-27
Set up keyboard based settings pages


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Android/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-Android's [Code of Conduct](https://github.com/scribe-org/Scribe-Android/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

Scribe-iOS provides settings for each individual keyboard. As seen on the [designs on Figma](https://www.figma.com/design/c8945w2iyoPYVhsqW7vRn6/scribe_public_designs?node-id=513-1616&t=J8vmgQqaca9JriVj-1) as well as in the image below, each keyboard that is activated has a settings page beneath the global settings tab.

![Keyboard Settings - Light](https://github.com/user-attachments/assets/c6b05fef-fef6-47d2-8f5e-cdda7694d4f4)

This issue would do the following:

- Each installed keyboard would have an option to go to its settings on the Settings tab
- The settings page for each keyboard would just have a back button for now
- If there are no keyboards installed, then an `Install keyboards` button would be present as seen in the designs [here](https://www.figma.com/design/c8945w2iyoPYVhsqW7vRn6/scribe_public_designs?node-id=653-1846&t=J8vmgQqaca9JriVj-1)
    - This is assuming that the user will not be forced to install a keyboard upon opening the app
    - We won't need to do this if we do insist that the user has already installed a keyboard to navigate through the application

### Contribution

Happy to support this any way that I can and review when we're ready to look into the solution! 😊

andrewtavis:
Assigning this to you per our discussions, @angrezichatterbox :)

angrezichatterbox:
Should I implement the Double space period  in the individual keyboard settings?

andrewtavis:
Hey @angrezichatterbox 👋

> Should I implement the Double space period in the individual keyboard settings? 

Yes, we decided that all keyboard based settings would be in individual keyboards for consistency as some do need to be there :)

> Also, Could I make the settings page similar to Scribe-iOS? This would include making the App Language go to the App Language Settings and the current checkboxes turned into a switch similar to Scribe-iOS and Figma. Should it be kept for a separate issue? 

I think it makes sense to include it in here, yes :) All sounds great 😊 Thank you!

angrezichatterbox:
[...]

PR #63, opened 2024-08-22
Implemented Language Based Settings 

angrezichatterbox:
### Contributor checklist

---

### Description
The pull request implemented Language-based Settings for the enabled language keyboards. I have also changed the min and target sdk and upgraded gradle to 8.5.2 to enable navigation features to app locale settings. I have also moved the files to the Android Studio-generated package. 

### Related issue

-   #46


andrewtavis:
All this is really great, @angrezichatterbox! One thing that I'm noticing is that the dark mode toggle is a bit weird 🤔 It's light mode when I first load the app, but then the toggle is technically on for dark mode. Then when I click it first nothing happens, then clicking it again it goes to the left, and then clicking it again it goes to the right. This is the only toggle that has this issue. Do you want to take a look and then we can merge this in after.

Thanks so much! 🚀

angrezichatterbox:
> All this is really great, @angrezichatterbox! One thing that I'm noticing is that the dark mode toggle is a bit weird 🤔 It's light mode when I first load the app, but then the toggle is technically on for dark mode. Then when I click it first nothing happens, then clicking it again it goes to the left, and then clicking it again it goes to the right. This is the only toggle that has this issue. Do you want to take a look and then we can merge this in after.
> 
> Thanks so much! 🚀

I have fixed the dark mode toggle. :)

andrewtavis (approved):
Thanks for the quick reaction to the feedback and another great contribution, @angrezichatterbox! Really is coming along now that we have two screens of the app looking like they do! I'll check in with you on next steps 😊

Issue #59, opened 2024-08-16
Add the Privacy policy page to About tab


angrezichatterbox:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Android/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-Android's [Code of Conduct](https://github.com/scribe-org/Scribe-Android/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue would add the Privacy Policy page to the About tab of the Scribe-Android application. The image of the page is given below and the designs can be checked out [here on Figma](https://www.figma.com/design/c8945w2iyoPYVhsqW7vRn6/scribe_public_designs?node-id=547-1500&t=C7GHrHdRubdKpcPw-1). 

<img src="https://github.com/user-attachments/assets/5c330651-0bac-45d0-9ef2-aa7ca1cb76b2" alt="Privacy Policy - Light" width="350"/>

- Note: The issue is currently blocked #58 where we would be adding the legal tab view to the about page where there would be a section to navigate to the privacy policy and third-party license page. 

### Contribution

I’m happy to take on this issue and am more than willing to assist anyone else who wants to work on it. 😄

Linfye:
I can work on this. Could you tell me more details about this?

angrezichatterbox:
Thank you very much!

The current About Page includes a Legal Tab with two options. This issue involves adding navigation from the Privacy Policy option to a new activity. The layout for this new activity is detailed in the issue description.

To address this issue, you need to:

- Create the new activity layout as specified in the issue description.
- Implement navigation so that selecting the Privacy Policy option in the Legal Tab takes the user to the new activity.
- You can achieve this by adding the new activity to the RecyclerView data in the About Activity.

A reference for a similar layout can be seen on the Wikimedia Scribe Page. 

Feel free to ask any queries :)

andrewtavis:
Realized that I forgot to assign, @Linfye :) Please let us know if you have any questions here!

andrewtavis:
Closed via #72 🚀 Thanks so much for the work of getting up to speed with Scribe-Android, @Linfye! Amazing to have your contributions here as well :)

PR #72, opened 2024-08-27
Add Privacy Policy page

Linfye:
### Contributor checklist

---

### Description

This PR adds a privacy policy page. 

But the page does not show line breaks. I have tried editing `string.xml` with "\n\n" but it changed back after I built the project. So it looks like this now:
PixPin_2024-08-27_16-13-13

I search it on Google and tried many ways but can't solve it. I am new in Android App developing. Can you help me with this? @angrezichatterbox
### Related issue

-   #59


andrewtavis:
Thanks for the PR, @Linfye! Greta to have you over here :) Let's be sure to not edit the texts as those are from Scribe-i18n and we need to be able to pull, but we can reset those after the PR. @angrezichatterbox, let us know if you have suggestions on the line breaks 😊

angrezichatterbox:
> Thanks for the PR, @Linfye! Greta to have you over here :) Let's be sure to not edit the texts as those are from Scribe-i18n and we need to be able to pull, but we can reset those after the PR. @angrezichatterbox, let us know if you have suggestions on the line breaks 😊

It seems that the script written for the Scribe-i18n adds a new line to the XML file instead of adding a \n. A possible fix would be to make the script to be ignoring the \n. It was an undersight from part. Sorry. Another fix could be using regex in the Kotlin front to load the text. It would be better to make the script ignore it right ?

app/src/main/res/layout/activity_privacy_policy.xml
```diff
@@ -0,0 +1,108 @@
...
+                    <LinearLayout
+                        android:layout_width="match_parent"
+                        android:layout_height="wrap_content"
+                        android:orientation="vertical"
+                        android:padding="16dp">
+
+                        <TextView
+                            android:singleLine="false"
+                            android:layout_width="304dp"
+                            android:layout_height="3560dp"
```
angrezichatterbox, line 69: Could this be changed to wrap_content. Since I am seeing a lot of space after the content since the layout height is fixed.

angrezichatterbox (commented):
Your PR looks mostly Great @Linfye. There is a small change to be made that's it. Hope we can see you make more contributions to Scribe-Android.

app/src/main/res/layout/activity_privacy_policy.xml
Linfye: I saved these space lines for the line breaks. It should fix then.

angrezichatterbox: Wouldn't wrap_content be enough even if the new lines are there?

Linfye: Oh, I know. I changed the `height`  to `wrap_content` now.

app/src/main/res/layout/activity_privacy_policy.xml
```diff
@@ -0,0 +1,107 @@
...
+                android:layout_width="match_parent"
+                android:layout_height="40dp" />
+
+            <TextView
+                android:id="@+id/textView"
+                android:layout_width="303dp"
+                android:layout_height="wrap_content"
+                android:layout_marginStart="16dp"
+                android:layout_marginTop="56dp"
+                android:text="@string/app.about.privacyPolicy"
```
angrezichatterbox, line 38: Could you change this text to say Keeping you safe from the strings.xml

angrezichatterbox:
@Linfye A temporary fix for the line breaks could be to use regex with HTML formatting. One thing I noticed is that most of the single lines should be in a single paragraph. You add a break like after each `. ` This could also be done  @andrewtavis.

app/src/main/res/layout/activity_privacy_policy.xml
Linfye: fixed this

Linfye:
> @Linfye A temporary fix for the line breaks could be to use regex with HTML formatting. One thing I noticed is that most of the single lines should be in a single paragraph. You add a break like after each `. ` This could also be done @andrewtavis.

But this will split these places:
image

🧐

andrewtavis:
Big thing is that we want the texts that are used here to be directly in Scribe-i18n. No formatting should be necessary within Scribe-Android. I'll check the changes in the most recent i18n PR and maybe we can bring those in here :)

andrewtavis:
Ok @Linfye and @angrezichatterbox: This branch now has the updated i18n keys for Android. Let me know if it's working!

Linfye:
> Ok @Linfye and @angrezichatterbox: This branch now has the updated i18n keys for Android. Let me know if it's working!

I pull the update and it works fine now. Thank you!😄

andrewtavis:
Ok great, @Linfye, but what were the changes to the string files in your last commit? These files should only be what is brought in from the Scribe-i18n project, so it changes are needed still then we need to make them in Scribe-i18n and then bring them here :) 

Maybe I'm misunderstanding though. Thanks so much for the work here!

Linfye:
> Ok great, @Linfye, but what were the changes to the string files in your last commit? These files should only be what is brought in from the Scribe-i18n project, so it changes are needed still then we need to make them in Scribe-i18n and then bring them here :)
> 
> Maybe I'm misunderstanding though. Thanks so much for the work here!

I didn't push those changes. :)

andrewtavis:
Just checking, I'm talking [this commit here](https://github.com/scribe-org/Scribe-Android/commit/f44020269c46dd159e3a84a19b3b2044d249fbba) that comes after my commits to import Scribe-i18n. Maybe it's looking weird on GitHub mobile (I'm on the way to work and am trying to check this)? I'm just considering that those changes might be conflicts for future Scribe-i18n updates as we need to maintain the same git history as the Scribe-i18n repo.

angrezichatterbox:
> Just checking, I'm talking [this commit here](https://github.com/scribe-org/Scribe-Android/commit/f44020269c46dd159e3a84a19b3b2044d249fbba) that comes after my commits to import Scribe-i18n. Maybe it's looking weird on GitHub mobile (I'm on the way to work and am trying to check this)? I'm just considering that those changes might be conflicts for future Scribe-i18n updates as we need to maintain the same git history as the Scribe-i18n repo.

That's the change the gradle bought. The gradle changed the existing values of the Android project from the Scribe-i18n assets folder.

andrewtavis:
Ah ok, perfect :) Thanks so much for the explanation, @angrezichatterbox, and sorry for the confusion, @Linfye! I had a feeling I just wasn't seeing something 😇

I'll bring this in later today :) Let us know if there's another issue that's of interest, @Linfye!

andrewtavis (approved):
Thanks for all the work here, @Linfye! Really happy to have you on the project and also it's so great that you're already onto your next issue 😊

We've getting a real team here now with multiple people! 🎉

Issue #76, opened 2024-08-30
Add back button to app menu sub pages


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Android/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-Android's [Code of Conduct](https://github.com/scribe-org/Scribe-Android/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

One thing that's needed for Scribe-Android is back buttons for when a user navigates to a sub pages within the app's menu. The back button can be seen in the [designs on Figma](https://www.figma.com/design/c8945w2iyoPYVhsqW7vRn6/scribe_public_designs?node-id=547-1500&t=CFEipBvDSeXA8rp2-1) at the top left of each page that in a sub page of the main installation, settings and about pages.

### Contribution

Happy to review this and get to it eventually if need be! 😊

andrewtavis:
@angrezichatterbox, would this one make sense for you as it's a bit more low level for how the menu functions?

andrewtavis:
I guess a custom action bar such that we can include the formatting of the top of it to make the color the same as the app background color in dark mode and also have it be that there's no line/shadow between the action bar and the app background? If that sounds good to you :?

angrezichatterbox:
> Generally matching Scribe-iOS as much as possible would be good, but if you then that there are benefits to changes then let me know and we can talk them though 😊 Happy to have things over here also go over to iOS if we think they're improvements! I think the way the bottom tab bar icon for the current page gets a bit bigger is actually pretty nice and maybe something we want for iOS. As of now it's just the same size. What do you think? You could make an issue in iOS for it? :)


Yes, the material navigation bar highlights the current tab by either bolding the text or increasing its size. I too feel this feature would be a good addition to Scribe-iOS, as it provides a clearer indication of the selected tab compared to just a color change. I can create an issue for this in Scribe-iOS. :).

angrezichatterbox:
I have created a pull request for this issue. I have tried to make it similar to Scribe-iOS. If there are any changes to be made I will be happy to do it :).

andrewtavis:
Closed by #77 🚀 We're getting really close with the menu, @angrezichatterbox! Crazy all of the progress we've been making over the last few months. Thanks so much!

PR #77, opened 2024-09-03
feat:Implemented back button in the action bar

angrezichatterbox:
### Contributor checklist

---

### Description

- Implemented the Back Button in the action bar 
- Made the Action Bar similar to Scribe-iOS

### Related issue

-   #76


andrewtavis:
This is looking so good, @angrezichatterbox! It's really looking like Scribe-iOS finally 🤩 One thing to note: Would it be possible for the action bar to have a bit of a margin at the top? Maybe bring everything down like a centimeter or whatever looks ok on your end? The headers seem a bit high, and the back buttons like of blend in with the icons at the top of the screen.

After this we'll be good to merge! 😊

angrezichatterbox:
> This is looking so good, @angrezichatterbox! It's really looking like Scribe-iOS finally 🤩 One thing to note: Would it be possible for the action bar to have a bit of a margin at the top? Maybe bring everything down like a centimeter or whatever looks ok on your end? The headers seem a bit high, and the back buttons like of blend in with the icons at the top of the screen.
> 
> After this we'll be good to merge! 😊



Screenshot 2024-09-04 at 9 03 50 AM
Screenshot 2024-09-04 at 9 10 10 AM

How does this look? Should I make them come slightly below in the case of the subpages?

andrewtavis:
Is looking great, @angrezichatterbox! The only thing I'd change is that sub headers like `How we work together` should be move to the right a bit so they're inline with the page header and the back button, and the back button in your screens should be black as it's light mode, so maybe check that the colors for those are being applied correctly 😊

andrewtavis (approved):
All looking really great, @angrezichatterbox! Really is coming along so well :) Just need the logo and the directions for the installation to be a bit more explicit, and then it'll really be looking like Scribe 😊

Issue #95, opened 2024-09-08
Convert all strings to Scribe-i18n keys


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Android/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-Android's [Code of Conduct](https://github.com/scribe-org/Scribe-Android/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

As discussed in #94, changes in [Scribe-i18n](https://github.com/scribe-org/Scribe-i18n) that just went through mean that we now have all the strings from the designs within the language i18n files. It would be great if we could update the Scribe-i18n subtree and then convert all strings in the application over to Scribe-i18n keys.

### Contribution

Happy to review and support as needed!

andrewtavis:
@angrezichatterbox, I'm also thinking that this one might make sense for you as this would allow you to check that I've added all of the strings you mapped out? Thoughts on picking this up? :)

angrezichatterbox:
> @angrezichatterbox, I'm also thinking that this one might make sense for you as this would allow you to check that I've added all of the strings you mapped out? Thoughts on picking this up? :)

Yes I'll work on this once #89 is done.

andrewtavis:
Thanks, @angrezichatterbox! We'll be in a great place once this and #89 are done :) :)

angrezichatterbox:
Should I run the scripts for the conversion of the i18n strings to XML now or should it be done in the i18n and then pushed there as the current XML isn't updated according to the updated jsons.

andrewtavis:
I think it's fine for you to run the scripts and send along a PR. I'll merge tonight and then work can begin on this tomorrow :)

andrewtavis:
https://github.com/scribe-org/Scribe-i18n/commit/5701774cc4f4cd7665d86eb264a176c9627b95d9 just sent along the changes you suggested for Scribe-i18n, @angrezichatterbox :) Maybe we can close this issue via another minor edit to convert the last missing strings?

andrewtavis:
Closed by #105 and #106 🚀 Thanks so much for this, @angrezichatterbox! Really awesome to be building so sustainably from the start 😊

PR #105, opened 2024-09-12
Reference all strings from scribe i18n

angrezichatterbox:
### Contributor checklist

---

### Description

- The PR adds reference to all strings from Scribe-i18n for Scribe-Android

### Related issue

-   #95 

### Additional information 

- I ran the script update_i18n_keys.sh for updating the keys. I hope I would not have caused any mess-ups in the git subtree. 
- The strings used in the user interfaces of Scribe-Android to be added to Scribe-i18n are:

"Show a popup on keypress",
"Dark mode"
“Enable Keyboard”
 <string name="third_party_licenses_body">The Scribe developers (SCRIBE) built the Android application "Scribe - Language Keyboards" (SERVICE) using third party code. All source code used in the creation of this SERVICE comes from sources that allow its full use in the manner done so by the SERVICE. This section lists the source code on which the SERVICE was based as well as the coinciding licenses of each.\n\nThe following is a list of all used source code, the main author or authors of the code, the license under which it was released at time of usage, and a link to the license.\n\n1. Simple Keyboard\n• Author:Simple Mobile Tools\n• License: GPL-3.0\n• Link: [https://github.com/SimpleMobileTools/Simple-Keyboard/blob/main/LICENSE</string](https://github.com/SimpleMobileTools/Simple-Keyboard/blob/main/LICENSE%3C/string)>


andrewtavis:
"Enable keyboard" won't be added to the strings, by the way. We'll make an issue soon to expand the directions with keys I already added.

andrewtavis (approved):
Looks great, @angrezichatterbox! I'll do the other strings now and then we can do one more PR to close out the issue :)

PR #106, opened 2024-09-13
Reference all strings from scribe i18n

angrezichatterbox:
### Contributor checklist

---

### Description

- The pull request adds the leftover strings from i18n to the app.
- It also corrects some layout issues in some pages which I could notice. 

### Related issue

-   #95


andrewtavis (approved):
Awesome work, @angrezichatterbox! We'll be able to bring in the other strings for other languages and any other corrections that may be on the way! So great that localization is supported so early in the process 😊

andrewtavis:
Sounds good, @angrezichatterbox! I can make some more issues for the keyboard soon :)

andrewtavis:
Tons of new issues just made, @angrezichatterbox! Many of the new ones are for the keyboard as well :) Let me know what sounds appealing, and I've also already pinged you in the ones I think would be good to look into first 😊

Issue #113, opened 2024-09-14
Expand keyboard installation instructions on installation screen


andrewtavis:
### Terms

- [X] I have searched [open and closed design issues](https://github.com/scribe-org/Scribe-Android/issues?q=is%3Aissue+label%3Adesign)
- [X] I agree to follow Scribe-Android's [Code of Conduct](https://github.com/scribe-org/Scribe-Android/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

As of writing the `Installation` screen of the application has `ENABLE KEYBOARD` as the title of the button to select the keyboard. The current state is shown below:

<img width="452" alt="Screenshot 2024-09-14 at 19 42 10" src="https://github.com/user-attachments/assets/9c7ff16f-73e4-4e44-9a68-908995fe850a">

We'd like to expand this with more detailed directions for the user as seen on the [designs on Figma](https://www.figma.com/design/c8945w2iyoPYVhsqW7vRn6/scribe_public_designs?node-id=458-1320&t=J8vmgQqaca9JriVj-1) and below:

![Keyboard Installation - Light](https://github.com/user-attachments/assets/9126c11a-5904-4181-971f-f5aa79ea0733)

The strings needed for this are:
- `app.installation.keyboard.keyboard_settings`
- `app.installation.keyboard.text_2` (we don't need `test_1` for Android)
- `app.installation.keyboard.text_3`
- The Android keyboard icon would go here
- `app.installation.keyboard.text_4`

Included in this would also be adding in the settings cog icon on the upper right corner of the option as is done in the privacy policy page and third party likenesses :)

### Contribution

Happy to discuss this with someone and review or also get to it myself eventually 😊

andrewtavis:
@Linfye, I'm thinking that this could be a pretty straightforward issue for you given your work to add in the other pages with the upper right icons? Would you like to work on this one?

andrewtavis:
Looking into #119 at the same time might also make sense, as the functionality of the button will change from enabling a keyboard to selecting keyboards to install, which is handled in #119 :) :)

Linfye:
I’m happy to do this, but I’m a bit busy at the moment. I might only be able to work on Scribe in about two weeks, is that okay?

andrewtavis:
100% ok, @Linfye :) I won't assign you for now so it's not on your to do list, but please let us know when you have more availability!

Linfye:
@andrewtavis I am working on this. You can assign me this now. 😊

[...]

PR #140, opened 2024-09-22
Add Installation Instructions 

Linfye:
### Contributor checklist

---

### Description
The current page looks like this:
image
And I have 3 questions:

1. Are there any click events that need to be added to this page? Currently, my PR has turned this page into a static page.
2. Currently, there is a 'Scribe' title in the top-left corner of the page, which is not present in the Figma design. I'm not sure how to remove it, could you give me some tips?
3. In the Figma design, "Open Scribe Settings" has an underline, and I don't know how to add it in the XML file. Could you help me with that?

Thanks. :)

### Related issue

-   #113


andrewtavis:
Hey @Linfye! This is looking amazing 🤩 Really can't wait to bring this in as the app menu will then be at a great point :)

A couple comments on some things that need to change before we merge:

The texts for the field should be:

- `app.installation.keyboard.keyboard_settings`
- `app.installation.keyboard.text_2` (we don't need test_1 for Android)
- `app.installation.keyboard.text_3`

So the final text should be:

```
1. Open keyboard settings
2. Activate keyboards that you want to use
3. When typing, press [KEYBOARD_ICON] to select keyboards
``` 

Specifically we don't want 🌐, but rather ⌨️ and the icon that's used on Android for switching keyboard input methods. On iOS we have the globe for this, but for Android there's a keyboard icon.

Answers to your questions:

1. Are there any click events that need to be added to this page? Currently, my PR has turned this page into a static page.

    - I think all's fine. We can add some toasts for the various buttons, but aside from that all's fine :)

3. Currently, there is a 'Scribe' title in the top-left corner of the page, which is not present in the Figma design. I'm not sure how to remove it, could you give me some tips?

    - @angrezichatterbox, do you have an idea on this?

4. In the Figma design, "Open Scribe Settings" has an underline, and I don't know how to add it in the XML file. Could you help me with that?

    - I'm not too sure on this, but let's not worry about it too much as the XML file will be changing in the comings months when we start converting the app layouts to jetpack compose. No need to worry for now :)

angrezichatterbox:
> > 3\. Currently, there is a 'Scribe' title in the top-left corner of the page, which is not present in the Figma design. I'm not sure how to remove it, could you give me some tips?
> 
> * @angrezichatterbox, do you have an idea on this?

A possible solution would be to hide the action bar on Creation of the MainActivity using supportActionBar.hide() and view pager changes would show the action bar and hide according to which page the user would be. There might be additional changes to be made but this could work.

Linfye:
Thank you for your responses. I have made the changes. Could you review it again? 😊

andrewtavis:
Sorry to ask for two final changes, @Linfye :)

We also want the keyboard installation field to be a button, which is why `Open keyboard settings` is blue and eventually underlined. Could you add an action to that field such that when you press it it opens the list of keyboards to install? So when we click anywhere on the keyboard installation instructions we should go to this page of the settings:

Screenshot_20240923_081747

And the last thing - also sorry for not being clear with this - can we get the ⌨️ changed to the icon that Android uses for switching keyboard inputs? Specifically the one below:

Screenshot 2024-09-23 at 08 18 41

Really is looking great! 😊 So nice to see Scribe-Android like this finally :)

Linfye:
I've made those changes you mentioned. You can review again now. Thanks. 😊

angrezichatterbox:
@Linfye Could u show the action bar in case of the other pages using supportActionBar?.show() command when the view pager changes it position. If it causes adjustment issues in case of nav bar changing position you could try modifying the xml layout of main activity to fill system window.

Linfye:
@angrezichatterbox, I'm not quite clear about what you mean. Are you suggesting to turn on the supportActionBar on the fragment_main.xml? Could you explain in detail how to do it? 😊

angrezichatterbox:
> @angrezichatterbox, I'm not quite clear about what you mean. Are you suggesting to turn on the supportActionBar on the fragment_main.xml? Could you explain in detail how to do it? 😊

Sorry I could have been more clear. 

So currently you have hidden the supportActionBar on the MainFragment. The action bar is not visible for the other pages as well. A possible fix for this would be to hide the support Action bar onCreate in the MainActivity. 

and then in the Main Activity in the view pager show the action bar for the other pages. 

https://github.com/scribe-org/Scribe-Android/blob/b1260afe6136a37c49caec534538d97e8617f0b7/app/src/main/java/be/scri/activities/MainActivity.kt#L84-L114

This might  cause the nav bar something to move up and down as the action bar is being changed continuously a possible fix for this would be to fill in window size for the relative layout of the main activity and then accordingly set the background color as well.

Linfye:
I tried several methods, but the action bar always has a strange up and down issue. In the end, I changed the layout of the Installation page and removed the title "Scribe" from the strings.xml characters. Currently, the page looks like this.

image

And action bar from other pages can display as expected.

Linfye:
> I'll let @angrezichatterbox speak to the action bar, @Linfye :) For the icon I was hoping that Android had a pre-built one we could use, but it looks like there's not one. I went ahead and made it :) Can you use the SVGs I'm attaching for light mode and dark mode?
> 
> android_keyboard_icon_dark android_keyboard_icon_light
> 
> Thanks so much!

OK, done. Thanks for the images! 😊

andrewtavis:
Let us know what the final review on your end is, @angrezichatterbox!

angrezichatterbox (approved):
The PR looks awesome @Linfye. The action bar works fine and is present for all pages.

andrewtavis:
Will play around with it a bit more after work and send along the final review, @Linfye! Thanks :)

andrewtavis (approved):
All's looking really good, @Linfye! I'll bring this in and note a few more things that need to be done in a new issue 😊 So awesome to see Scribe-Android like this :D

PR #152, opened 2024-09-24
fix:Fixed crash due to tableName being changed to lowercase

angrezichatterbox:
### Contributor checklist

---

### Description

- The issue fixes the app not building due to tableName in Clips.kt changed to lowercase.


andrewtavis (approved):
Thanks for the detailed review leading to this getting fixed, @angrezichatterbox! Hope all's well with classes so far :)

Issue #137, opened 2024-09-21
Switch text proxy to the command bar when it is shown


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Android/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-Android's [Code of Conduct](https://github.com/scribe-org/Scribe-Android/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

One of the main features of Scribe is that the user can type commands into the keyboard. The first step for this is allowing for the user to type into the command bar whenever it is shown. The behavior is:

- The user types normally when the keyboard state is idle
- If the user presses the Scribe key and then selects a command, then the command bar is shown (done in #89)
- When the command bar is shown, when the user types they should be typing into the command bar, not the application where they're using the keyboard

### Contribution

Happy to support with this and review as needed, or can also get to it myself eventually! This would be a major milestone for Scribe-Android 😊

angrezichatterbox:
I would like to work on this issue.

andrewtavis:
Nice @angrezichatterbox! Would be also super great to pick this one up :) Do have the ability to self assign issues with triage rights? You're more than welcome to self assign and let me know that you did, just in case we need to discuss the issue as you work on it :)

angrezichatterbox:
> Nice @angrezichatterbox! Would be also super great to pick this one up :) Do have the ability to self assign issues with triage rights? You're more than welcome to self assign and let me know that you did, just in case we need to discuss the issue as you work on it :)

Yes I do have access to self assign. I just wanted to let you know before I assign myself. I would assign myself and let you know from the next :)

andrewtavis:
Looking forward, and thanks for taking the initiative!

angrezichatterbox:
In Scribe-iOS I could see that the enter key changes its key symbol when translate, conjugate, or plural mode. And when we click on it it's getting added to the input field from the command bar and then it goes back to the usual keyboard with the scribe key. Can I assume that once the enter is clicked in the translate conjugate or plural state it goes back to idle mode?

andrewtavis:
Yes, exactly, @angrezichatterbox! And adding the functionality you described might also be nice, but we can also make new issues for things like the enter key switching to a play icon, etc. Let me know what you think!

I definitely would write an enum for all of the states and use that going forward. When we didn't have that state enum it was quite annoying to track...

[...]

PR #159, opened 2024-09-30
Implement switch text proxy to the command bar when shown 

angrezichatterbox:
### Contributor checklist

---

### Description

The pull request implements switching the text proxy to the command bar when it is shown. In specific there is a switch for all states except the select command state and idle state.

### Related issue

-   #137


andrewtavis (approved):
All looking good here, @angrezichatterbox! I'll bring this in and make the other issue for switching the icon :)

Issue #144, opened 2024-09-23
Remove right margin from the card view present in the Privacy Policy and Third Party Licenses page 


angrezichatterbox:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Android/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-Android's [Code of Conduct](https://github.com/scribe-org/Scribe-Android/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

Currently the third party license page and the privacy policy page card view has a right margin. This causes the page to have an empty space towards the right. The issue is to remove the right margin so that the user interface would be much better. 

<img width="200" alt="Screenshot 2024-09-23 at 10 06 46 AM" src="https://github.com/user-attachments/assets/1047b0ac-2462-4f44-b860-aa4b998f4faf">

<img width="200" alt="Screenshot 2024-09-23 at 10 06 58 AM" src="https://github.com/user-attachments/assets/63b2218e-f945-4544-97e1-e51acbea1ee2">

As seen from above the texts don't go towards the right even though there is space. We have to remove that space from the right in the individual card view. 

### Contribution

I would love to work on this and am more than happy to help anyone fix this issue :)

andrewtavis:
Closed by #168 🚀 Thanks, @henrikth93! 🥳

PR #168, opened 2024-10-02
Henrikth93 android margin #144

henrikth93:
### Contributor checklist

---

### Related issue

-   #144

### DIFF
--- a/app/src/main/res/layout/fragment_privacy_policy.xml
+++ b/app/src/main/res/layout/fragment_privacy_policy.xml
@@ -45,11 +45,14 @@
                         android:padding="16dp">
 
                         <TextView
-                            android:layout_width="304dp"
+                            android:id="@+id/privacyPolicyTextView"
+                            android:layout_width="match_parent"
                             android:layout_height="wrap_content"
+                            android:layout_marginTop="16dp"
                             android:layout_gravity=""
                             android:text="@string/app.about.legal.privacy_policy.text"
-                            android:textSize="16sp" />
+                            android:textSize="16sp"
+                            android:layout_marginEnd="6dp"/>
 
                     </LinearLayout>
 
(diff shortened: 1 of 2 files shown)


angrezichatterbox (commented):
@henrikth93 Thank you very much for the PR.  

Could the text View margins be changed instead of card view, or could the text view margins be changed to match the parent so that we would cover the entire width of the parent, which is card view. We would, however, then have to find ways to adjust so that the text does not go under the corner image.

henrikth93:
> @henrikth93 Thank you very much for the PR.
> 
> Could the text View margins be changed instead of card view, or could the text view margins be changed to match the parent so that we would cover the entire width of the parent, which is card view. We would, however, then have to find ways to adjust so that the text does not go under the corner image.

I am thinking if one way to do it is to split the original texts into two parts, one parts which is adjacent to the corner image, and has different layout settings to prevent them from overlapping, and a different text view which fills out the whole box

angrezichatterbox:
> > @henrikth93 Thank you very much for the PR.
> > Could the text View margins be changed instead of card view, or could the text view margins be changed to match the parent so that we would cover the entire width of the parent, which is card view. We would, however, then have to find ways to adjust so that the text does not go under the corner image.
> 
> I am thinking if one way to do it is to split the original texts into two parts, one parts which is adjacent to the corner image, and has different layout settings to prevent them from overlapping, and a different text view which fills out the whole box

Yes, That would work fine as well. But it would depend on how the text is presented for now. If all the text is present together, we could try passing from Kotlin code where we would split the text and then pass to the text views we have in the XML.

andrewtavis:
Let's make sure to remember that the texts need to be in one string for Scribe-i18n :) Anything that generally works is fine. Honestly maybe a redesign to get rid of those corner triangles would be good, but that's for maybe post Outreachy :)

angrezichatterbox:
@henrikth93 It looks good now. What do you think would it  be better if the card view margin end is changed to 8 dp or 9 dp?

henrikth93:
> @henrikth93 It looks good now. The text does not get hidden in most languages . What do you think would it be better if the card view margin end is changed to 8 dp or 9 dp?

Changed it to 10.

angrezichatterbox:
It looks great to me now  Thanks :)

andrewtavis (approved):
Thanks for the changes, @henrikth93, and also for the review, @angrezichatterbox! Is looking really great 😊

Issue #145, opened 2024-09-23
Version needs to be converted to an external link to directs towards the releases on GitHub


angrezichatterbox:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Android/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-Android's [Code of Conduct](https://github.com/scribe-org/Scribe-Android/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

The About page currently has a a version option under the Feedback and support Recycler view. The version option currently doesn't redirect to any page and has no action defined to it now. The issue aims to change the existing icon of the version page to showcase that it is to navigate to an external link as for the case of the see the code on GitHub option. and the option should be redirected to the Github Releases page of Scribe-Android.

Image references given below:

| Existing                                       | To be Changed to                                |
|-----------------------------------------------|------------------------------------------------|
| <img src="https://github.com/user-attachments/assets/2a41fed5-afb5-4e4f-a26c-2ae8cab63d90" width="389" /> | <img src="https://github.com/user-attachments/assets/09fb6202-7a50-4914-b883-01a994f5c703" width="389" /> |

Note: There are separate issues to add the divider. The issue is to change the version option third icon and add navigation to GitHub Releases Page

### Contribution

I would love to work on this and am more than happy to help anyone fix this issue :)

DeKaN:
Hi, can I take the issue?

andrewtavis:
By all means, @DeKaN! Looking forward to the contribution!

DeKaN:
@andrewtavis Can you provide an example of the url? Should it be `https://github.com/scribe-org/Scribe-Android/releases/` or `https://github.com/scribe-org/Scribe-Android/releases/tag/<VERSION>` ?

angrezichatterbox:
It should navigate to this URL https://github.com/scribe-org/Scribe-Android/releases/

andrewtavis:
Closed by #169 🚀 Thanks for the feature, @DeKaN! Hope we can work with you again :)

PR #169, opened 2024-10-03
Add version output to AboutFragment

DeKaN:
### Contributor checklist

---

### Description

Adds current version text to AboutFragment. Clicking on it opens Github Releases page. Version text is generated from property `android.defaultConfig.versionName` in build.gradle.kts

### Related issue

-   #145


app/src/main/assets/i18n/Scribe-i18n/values/en-US/string.xml
```diff
@@ -27,7 +27,7 @@
     <string name="app.about.feedback.rate_conjugate">Rate Scribe Conjugate</string>
     <string name="app.about.feedback.rate_scribe">Rate Scribe</string>
     <string name="app.about.feedback.title">Feedback and support</string>
-    <string name="app.about.feedback.version">Version</string>
+    <string name="app.about.feedback.version">Version %1$s</string>
```
DeKaN, line 30: Should be updated in Scribe-i18n repo also

angrezichatterbox: Could you remove these changes for now where the assets folder values are being changed? It would mess up the git subtree.  @andrewtavis Should we make the text for now be taken from strings.xml which Android has given by default until the string is changed?

angrezichatterbox (commented):
Thank you so much for the PR @DeKaN. Could u remove the changes from the assets and temporarily add the value to the strings.xml file only and then the PR would be perfect

DeKaN:
@angrezichatterbox Done

angrezichatterbox (commented):
Could you run the app ones so that changes to values-es would be retracted? Also, could you add the version string to strings.xml rather than string.xml? So it is like every time the app is run the values in string.xml files get changed.

DeKaN:
Yes, Gradle overwrites string.xml by version from assets (that's why I added changes there). 
If I move `Version` string to `strings.xml`, I have to use another resource id and the string will be excluded from your i18n. Are you sure?

angrezichatterbox:
> Yes, Gradle overwrites string.xml by version from assets (that's why I added changes there). If I move `Version` string to `strings.xml`, I have to use another resource id and the string will be excluded from your i18n. Are you sure?

The reference remains the same you could call using R.string.name_of_string or @string/string_name

DeKaN:
Resources must have unique ids. So string with id `app.about.feedback.version` must be removed from all `string.xml` files, if you want to put it to `strings.xml`. But you don't want to change files in `assets`. 
I can add it with another id, but you'll get this conflict later when you update the id in i18n repo

angrezichatterbox:
> Resources must have unique ids. So string with id `app.about.feedback.version` must be removed from all `string.xml` files, if you want to put it to `strings.xml`. But you don't want to change files in `assets`. I can add it with another id, but you'll get this conflict later when you update the id in i18n repo

I forgot that there was a Version string. For not just add a new string to strings.xml with the resource id of your choice and then use the string.

DeKaN:
Done

andrewtavis (approved):
Thanks for the changes, @DeKaN, and also for your two's collaboration 😊 Just tested and it's working fine. We can figure out the specifics of the text once we bring the Scribe-i18n subtree in next time :)

Issue #122, opened 2024-09-14
Add dividers between the autosuggestion buttons


andrewtavis:
### Terms

- [X] I have searched [open and closed design issues](https://github.com/scribe-org/Scribe-Android/issues?q=is%3Aissue+label%3Adesign)
- [X] I agree to follow Scribe-Android's [Code of Conduct](https://github.com/scribe-org/Scribe-Android/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue would add two very thin dividers between the autosuggestions as is common on system keyboards. The current keyboard interface is:

<img width="470" alt="Screenshot 2024-09-14 at 20 03 43" src="https://github.com/user-attachments/assets/cfa9a12a-4808-430f-9cdb-d539f6bea704">

There should be a 1px horizontal line between each of the autosuggestion keys that is 80% the height of the Scribe key and be centered vertically to the Scribe key and autosuggestion options. The color of these dividers should be `Special key color`. The iOS UI can be seen in the [App Store images on Figma](https://www.figma.com/design/c8945w2iyoPYVhsqW7vRn6/scribe_public_designs?node-id=403-9450&t=u8w7568Wwpzvvg5t-1).

### Contribution

Happy to support with this as needed and review once it's ready! Also happy to get to this myself eventually if needed 😊

andrewtavis:
Note that this issue is related to #121 where the special keys will have their colors set :)

catreedle:
Hi, could you please let me know which files are related to this issue?

andrewtavis:
@angrezichatterbox, could you assist here? :)

andrewtavis:
Closed by #179 🚀 Appreciate all the contributions, @catreedle :) :)

PR #179, opened 2024-10-09
add dividers between autosuggestion #122

catreedle:
### Contributor checklist

---

### Description

This PR adds dividers between the autosuggestion buttons.
divide
 Since the separator already exists, I made minor adjustments to the [keyboard_view_command_options.xml](https://github.com/scribe-org/Scribe-Android/blob/main/app/src/main/res/layout/keyboard_view_command_options.xml) file.

Please let me know if this approach is incorrect or if there are other considerations I should keep in mind. I look forward to your feedback! :)

Thank you!

### Related issue

-   #122


andrewtavis:
This is generally looking great, @catreedle, but one thing would be that on dark mode they should be a bit more grey :) We can do the color of the special keys for now. I think this will be good to merge after that!

catreedle:
You're right! I didn’t notice the color was off. I thought dark mode already looked different, but it turns out it’s just the contrast 😅. I’ll do my best to fix it

andrewtavis (approved):
Thanks so much for the quick fix and the continued great contributions, @catreedle! 😊

Issue #125, opened 2024-09-18
Fix linting issues due to Large Classes 


angrezichatterbox:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Android/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-Android's [Code of Conduct](https://github.com/scribe-org/Scribe-Android/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

The LargeClass rule is currently disabled. This rule helps in detecting classes that are too large, which can lead to code that is harder to maintain. Re-enabling this rule will aid in managing class complexity.

Steps to see which lines have to be modified:

1. Comment or set to true the following line in detekt.yml:

```
LargeClass:
    active: true

```

2. Run the command `./gradlew detekt`.
3. Review the results to identify classes that need refactoring.

### Contribution

I would love to work on this issue and am more than willing to help anyone fix this issue. :)

andrewtavis:
Closed by #185 :) Thanks, @KesharwaniArpita!

PR #185, opened 2024-10-10
Update detekt.yml to fix the large class lint issue

KesharwaniArpita:
### Contributor checklist

---

### Description

This PR fixes linting issues due to Large Class Block by removing the line of code:
```
EmptyCatchBlock:
        active: true
```

### Related issue

-   #125

### DIFF
--- a/app/src/main/java/be/scri/views/MyKeyboardView.kt
+++ b/app/src/main/java/be/scri/views/MyKeyboardView.kt
@@ -57,6 +57,7 @@ import be.scri.helpers.SHIFT_ON_PERMANENT
 import java.util.Arrays
 
 @SuppressLint("UseCompatLoadingForDrawables")
+@Suppress("LargeClass")
 class MyKeyboardView
     @JvmOverloads
     constructor(
--- a/detekt.yml
+++ b/detekt.yml
@@ -29,8 +29,6 @@ complexity:
         active: false
     CyclomaticComplexMethod:
         active: false
-    LargeClass:
-        active: false
 
 potential-bugs:
     ImplicitDefaultLocale:


andrewtavis:
As with the other one, @KesharwaniArpita, it's not just to remove the line, but also to fix the issues that are being hidden by it. Looks like this one is just one issue, but then [splitting MyKeyboardView](https://github.com/scribe-org/Scribe-Android/actions/runs/11279286899/job/31376184121?pr=185) is likely going to be really difficult 🤔

KesharwaniArpita:
Surely @andrewtavis , I'll do better

angrezichatterbox:
> As with the other one, @KesharwaniArpita, it's not just to remove the line, but also to fix the issues that are being hidden by it. Looks like this one is just one issue, but then [splitting MyKeyboardView](https://github.com/scribe-org/Scribe-Android/actions/runs/11279286899/job/31376184121?pr=185) is likely going to be difficult 🤔

Yes, you're right @andrewtavis. It would not be easy to split the class and would require a major refactoring which might include several changes to the overall codebase.

andrewtavis:
Is there a way for us to put an ignore comment to turn `detekt` off for this file such that the tests pass, and then from there we'll have this fail if another class gets to be too large? We can then pick this up again at some point for a more full refactor? It looks like [this discussion here](https://github.com/detekt/detekt/discussions/4825) might have some ways of doing this, i.e. adding something like `excludes: ['/Scribe-Android/app/src/main/java/be/scri/views/MyKeyboardView.kt']`?

KesharwaniArpita:
> Is there a way for us to put an ignore comment to turn `detekt` off for this file such that the tests pass, and then from there we'll have this fail if another class gets to be too large? We can then pick this up again at some point for a more full refactor? It looks like [this discussion here](https://github.com/detekt/detekt/discussions/4825) might have some ways of doing this, i.e. adding something like `excludes: ['/Scribe-Android/app/src/main/java/be/scri/views/MyKeyboardView.kt']`?

You mean by that instead of going straight to this file, we first focus on refactoring other files first and then move back to it later? Like a bottom to top approach?

andrewtavis:
This is the only class that has this problem, so if we ignore it or its file and just have that class in the file, then we'll be fine with this for now :)

angrezichatterbox:
> Is there a way for us to put an ignore comment to turn `detekt` off for this file such that the tests pass, and then from there we'll have this fail if another class gets to be too large? We can then pick this up again at some point for a more full refactor? It looks like [this discussion here](https://github.com/detekt/detekt/discussions/4825) might have some ways of doing this, i.e. adding something like `excludes: ['/Scribe-Android/app/src/main/java/be/scri/views/MyKeyboardView.kt']`?

@Suppress("LargeClass")

We can add this to the top of the class.

andrewtavis:
Do you want to make the change here, @KesharwaniArpita? When the tests pass we can then close this :)

andrewtavis:
Feel free to commit the change to this PR, @KesharwaniArpita :) It's basically adding `@Suppress("LargeClass")` right above [this class right here](https://github.com/scribe-org/Scribe-Android/blob/efc20efb00ae57dc50093377b9e01e38fd386fd8/app/src/main/java/be/scri/views/MyKeyboardView.kt#L60) and then we'll be done here 😊

KesharwaniArpita:
Sure!! Thanks @andrewtavis and @angrezichatterbox

andrewtavis (approved):
Fantastic, @KesharwaniArpita, and thanks for the great suggestion, @angrezichatterbox! 😊

Issue #165, opened 2024-10-02
Rate Scribe needs an action


angrezichatterbox:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Android/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-Android's [Code of Conduct](https://github.com/scribe-org/Scribe-Android/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

The Rate Scribe Option in the About Page currently doesn't have an action. The option has been implemented to be rated in either F-Droid or in the Google Play store depending on where it is being installed from. 

### Contribution

I would love to work on this and am more than willing to help anyone solve this issue.

MostafaMohamed2002:
can we do it before you upload the app to playstore ??

angrezichatterbox:
I am not sure about F-Droid. We could create the function for it as we can use the app identifier for the Play Store URL that would not change on the app release. I am not about how it would function before release.

henrikth93:
I have done some on this. Gonna contact you later @angrezichatterbox

andrewtavis:
Ya for this maybe we can comment out the code for the the actual Play store and F-Droid actions and leave the toast for now, but still have the code that we need once we have the URLs :)

andrewtavis:
I guess we can close for now and do another issue later when we have the URLs :) Closed in #187. Thanks @henrikth93!

PR #187, opened 2024-10-11
Add action for "rate scribe"

henrikth93:
This commit will add an action for "rate scribe"

### Contributor checklist

---

### Related issue

-   #165

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -114,6 +114,7 @@ dependencies {
     implementation("androidx.recyclerview:recyclerview:1.3.2")
     implementation("androidx.cardview:cardview:1.0.0")
     implementation("androidx.viewpager2:viewpager2:1.1.0")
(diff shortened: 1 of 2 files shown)


angrezichatterbox (commented):
Great PR, @henrikth93. This is one method to do it and is ideal since we also have F-droid. Also, I don't think we need to comment on those lines, as the statements already handle show no installation source when not installed from F-droid or Play Store. 


However, for the Play Store, an ideal method would be to use the Play Store API. It has an advantage. It would show a dialogue for rating once the app is on the Play Store, and for now, when used, the review would always return successfull.

angrezichatterbox (commented):
Look good to me now :) Thanks @henrikth93

andrewtavis (approved):
Wonderful stuff, @henrikth93, and thanks for the review and directing, @angrezichatterbox! 🚀🚀

Issue #111, opened 2024-09-14
Fix linting issue due to Long Methods


angrezichatterbox:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Android/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-Android's [Code of Conduct](https://github.com/scribe-org/Scribe-Android/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

The LongMethod rule is currently inactive. This rule identifies methods that are too long, which can affect readability and maintainability. Re-enabling this rule will help in managing method size and improving code quality.

Steps to see which lines have to be modified:

- Comment or set to true the following line in detekt.yml:
```
LongMethod:
    active: true

```
- Execute the command ./gradlew detekt.
- Check the output to find methods that need to be refactored.

### Contribution

I would love to work on this issue and am more than willing to help anyone fix this issue :)

henrikth93:
I can take a look at this one

andrewtavis:
Thanks for picking it up, @henrikth93 :)

andrewtavis:
Closed by #202 :) Thank you, @henrikth93!

PR #202, opened 2024-10-15
Add suppressions

henrikth93:
This commit will add suppressions for "Long method"

### Contributor checklist

---

### Related issue

-   #111

### DIFF
--- a/app/src/main/java/be/scri/fragments/LanguageSettingsFragment.kt
+++ b/app/src/main/java/be/scri/fragments/LanguageSettingsFragment.kt
@@ -19,6 +19,7 @@ import be.scri.databinding.FragmentLanguageSettingsBinding
 import be.scri.helpers.CustomAdapter
 import be.scri.models.SwitchItem
 
+@Suppress("LongMethod")
 class LanguageSettingsFragment : Fragment() {
     private var _binding: FragmentLanguageSettingsBinding? = null
     val binding get() = _binding!!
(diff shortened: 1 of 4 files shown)


andrewtavis:
@angrezichatterbox, are you good with the suppressions here, or should any of these be broken up?

angrezichatterbox:
> @angrezichatterbox, are you good with the suppressions here, or should any of these be broken up?

I suggested using Suppress for now as some of the classes would be modified or deleted. @henrikth93 Apologies for earlier stating that MyKeyboardView has to be suppressed for handling LongMethod. We could refractor a few of its functions but we would have to consider a throughout check of whether the keyboard is being drawn properly. Other classes for now I think its fine to have them suppressed as it would be removed later on.

andrewtavis:
Now I'm just confused 😅 Sorry, so we need to be explicit to say that we want tests to run? And the test was passing, then I removed the true entries because I thought it would default to being ran, and then the test failed, and now I've added them back in and it fails still?

angrezichatterbox:
> Now I'm just confused 😅 Sorry, so we need to be explicit to say that we want tests to run? And the test was passing, then I removed the true entries because I thought it would default to being ran, and then the test failed, and now I've added them back in and it fails still?

We have to rebase the branch with the main and then fix the remaining linting issues. This would fix the problem.  I had a similar issue with my PR where I was shown a linting issue in a line that was not present in my branch.

andrewtavis (approved):
Beautiful stuff, @henrikth93! 🚀 Thanks for the quick fix here and cleaning my mess :)

Issue #195, opened 2024-10-13
Add unit tests to PR Checks for Scribe Android


albendz:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Android/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-Android's [Code of Conduct](https://github.com/scribe-org/Scribe-Android/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

# Summary

Unit tests can be run with commands `./gradlew test` or `./gradlew check`. Note that the `check` task will also run other things like detekt. This issue tracks adding running these commands as part of PR checks.

This will be added to the `./github/workflows` folder with the other PR workflows.

One question is whether the unit tests should be merged with the ktlint PR workflow or kept separate.

### Contribution

I can work on this one. Feel free to assign to me with the recommendation on whether to merge this with the detekt PR workflow or have separate.

albendz:
I can pick this one up!

andrewtavis:
Thanks for picking it up, @albendz! :) :)

Linfye:
@andrewtavis Can this issue be closed by #209? 😊

andrewtavis:
Can be, yes @Linfye :) Thanks for the note and for the contributions, @albendz!

PR #209, opened 2024-10-18
feat: Add unit test check to PR workflows

albendz:
### Contributor checklist

---

### Description

Add PR workflow to run unit tests via the gradle test task.

Tested by running `./gradlew test` locally and making sure tests pass and fail as expected.

PR workflow will run on this PR.

### Related issue

-   #195

### DIFF
--- a/.github/workflows/pr_unit_test.yaml
+++ b/.github/workflows/pr_unit_test.yaml
@@ -6,7 +6,7 @@ on:
         types: [opened, reopened, synchronize]
 
 jobs:
-    ktlint:
+    unit-test:
         runs-on: ubuntu-latest
         name: Run unit tests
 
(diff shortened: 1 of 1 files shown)


.github/workflows/pr_unit_test.yaml
```diff
@@ -0,0 +1,27 @@
+name: pr_unit_test
+on:
+    pull_request:
+        branches:
+            - main
+        types: [opened, reopened, synchronize]
+
+jobs:
+    ktlint:
```
MostafaMohamed2002, line 9: change job name to something else because you can't have two jobs with the same name .

MostafaMohamed2002: check conflict here : https://github.com/scribe-org/Scribe-Android/blob/a5ce35e85e8c82a6c4442d7e31a1958733a3029c/.github/workflows/pr_ktlint_detekt.yaml#L9

andrewtavis (approved):
Checking this with @angrezichatterbox right now :) Thanks for the contribution, @albendz, and also for the review, @MostafaMohamed2002!

Issue #66, opened 2024-08-24
Add emoji autocompletions


angrezichatterbox:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Android/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-Android's [Code of Conduct](https://github.com/scribe-org/Scribe-Android/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

The issue is to implement emoji suggestions based on the text we would type using the keyboard. 

- The issue is currently being blocked by #64 where we would be adding a menu option to enable and disable emoji suggestions and #138 where the buttons are added

### Contribution

I’m happy to take on this issue and am more than willing to assist anyone else who wants to work on it. 😄

Linfye:
Is this issue ready? Can I start working on it now? 😊

andrewtavis:
We'd need #89 to get finished before we work on this one, @Linfye :) Do you want to pick up one of the refactoring issues, and hopefully @angrezichatterbox can finish up #89 in the coming days?

andrewtavis:
Assigning you to work on it when it's unblocked though :)

Linfye:
OK, I'll work on other issues now. 😊

angrezichatterbox:
> We'd need #89 to get finished before we work on this one, @Linfye :) Do you want to pick up one of the refactoring issues, and hopefully @angrezichatterbox can finish up #89 in the coming days?

Yes I will try my level best to complete it in the coming days.

andrewtavis:
As you have time for it, @angrezichatterbox :)

andrewtavis:
Hey @Linfye 👋 I just added in #138 to add in the buttons for emoji autosuggestions, which in hindsight is also blocking this (updated the issue text to reflect this). Would you have interest in adding these buttons to the keyboard interface?

Linfye:
I'd like to do this. I will start it after finishing #113. 😄

andrewtavis:
Nice @Linfye! Feel free to write in there and I'll assign :)

Linfye:
Is this issue ready now? @andrewtavis 😊

andrewtavis:
I'd say so, yes @Linfye :) Maybe we'll need a minor edit post data update, but I actually don't think that the emoji data will change much. Unblocking 🎉

andrewtavis:
[...]

PR #220, opened 2024-10-22
Add emoji autocompletions

Linfye:
### Contributor checklist

---

### Description
~~This is a draft PR, and the functionality is not yet complete. I will improve it following the steps below.~~

1. Load the external English language database into a hash table for easy querying.
2. Obtain the words the user is typing.
3. Retrieve the corresponding emoji from the hash table.
4. Display the emoji on the button.
5. Output the emoji when the button is clicked.
6. Implement these functionalities in SimpleKeyboard to extend support for other languages.

### Related issue

-   #66


Linfye:
@andrewtavis  I have completed this work, could you review the code? Thanks. 😊

There is still an small problem. After outputting an emoji using this feature, you need to press the backspace twice to delete the outputted emoji. 

This seems to be related to the encoding method of the emoji, and it should be possible to resolve this issue by modifying the `MyKeyboard.KEYCODE_DELETE` part in the `onKey` function. Or do you have any other suggestions? I can open an issue for this problem after this PR is completed.

angrezichatterbox (commented):
Thanks, @Linfye for this. Everything looks and functions great. 

And yes, I do think modifying the delete key function would be the right choice. We could add a condition where if the previous character is an emoji, it would delete the surrounding twice and the other cases as usual.

andrewtavis (approved):
Really appreciate you keeping at this for as long as you have, @Linfye 😊 So great to have you on the project, and looking forward to seeing this feature develop even more!

Issue #226, opened 2024-10-26
Deleting an Emoji Requires Two Backspaces


Linfye:
### Terms

- [X] I have searched all [open bug reports](https://github.com/scribe-org/Scribe-Android/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow Scribe-Android's [Code of Conduct](https://github.com/scribe-org/Scribe-Android/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

In the current version, deleting an emoji requires pressing backspace twice due to the way emojis are encoded.

### Device type

Medium Phone API 35 in Android Studio Emulator

### Versions

The latest main branch

Linfye:
I'd like to work on this. 😊

andrewtavis:
Thanks for the issue and the dedication to emojis, @Linfye! :)

andrewtavis:
Can you set the button color to transparent in the work for this one as well, @Linfye? Would be a nice step to getting it to a more final state :)

Linfye:
> Can you set the button color to transparent in the work for this one as well, @Linfye? Would be a nice step to getting it to a more final state :)

👌

andrewtavis:
Closed by #229 😊 Thanks for this, @Linfye! Great to have this functionality be so responsive and what the user would expect :)

PR #229, opened 2024-10-27
Fix the Button Deletion Issue

Linfye:
### Contributor checklist

---

### Description
I modified the `handleDelete` function for deleting the emoji and set the emoji button to semi-transparent.

### Related issue

-   #226


Linfye:
@andrewtavis Are you referring to setting the button color to semi-transparent? Initially, I tried setting it to fully transparent, but there was no feedback animation when using the button, which felt strange. So I chose semi-transparency. If you have any other requirements, I can make adjustments. 😊

andrewtavis (approved):
All's working fine, @Linfye :) We'll need to rework the way that the command area buttons are responding anyway, so this is an improvement for now. Ideally we'd have a different highlight animation that's similar to what we have in iOS, but this can be gotten to later 😊

Appreciate the continued support!

Issue #90, opened 2024-09-07
 Replace Magic Numbers with Named Constants as identified by detekt


angrezichatterbox:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Android/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-Android's [Code of Conduct](https://github.com/scribe-org/Scribe-Android/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

Linting issues have been identified by Detekt due to Magic Numbers across the codebase. Magic numbers are when numbers are used in certain code lines within the definition of what that number means. 

Example:

Before fixing:

```
fun isOverSpeeding(speed: Int): Boolean {
    return speed > 60  // Magic number 60 lacks clarity 
}

```

After fixing:

```
const val SPEED_LIMIT = 60  // Named constant provides context

fun isOverSpeeding(speed: Int): Boolean {
    return speed > SPEED_LIMIT
}

```

These can be identified across the code by the following steps:

- Enable the MagicNumber detection from the detekt.yml file. 
- Run the command ./gradlew detekt
- You would be able to see all the files where Linting issues are caused due to Magic Numbers

### Contribution

I would love to work on this and am more than willing to help anyone solve this issue. 😄

Linfye:
Could you assign this issue to me? The command returned around 500 errors, so resolving this issue might take some time.

Linfye:
Could you assign this issue to me? The command returned around 500 errors, so resolving this issue might take some time.

angrezichatterbox:
Yes, Feel free to work on this issue. If you have any questions, please don't hesitate to ask us. Do you think there is any need to break down this issue?

Linfye:
That's OK. I can work on this and I will ask questions when necessary. 😊

andrewtavis:
Thanks @Linfye and @angrezichatterbox! :)

andrewtavis:
Closed by #234 🚀 Thanks for getting to all of these issues recently, @Linfye! Amazing stuff :)

PR #234, opened 2024-10-30
Fix Magic Number Issues

Linfye:
### Contributor checklist

---

### Description
Thanks to the previous contributors for their refactoring work, which removed a lot of unnecessary code. I only need to modify about 50 formatting errors instead of the 500 I mentioned in #90. 😊

### Related issue

-   #90

### DIFF
--- a/app/src/main/java/be/scri/services/SimpleKeyboardIME.kt
+++ b/app/src/main/java/be/scri/services/SimpleKeyboardIME.kt
@@ -62,7 +62,7 @@ abstract class SimpleKeyboardIME(
     private var emojiSpaceTablet2: View? = null
     private var emojiBtnTablet3: Button? = null
 
-    // how quickly do we have to doubletap shift to enable permanent caps lock
+    // How quickly do we have to doubletap shift to enable permanent caps lock.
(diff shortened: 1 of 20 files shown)


app/src/main/java/be/scri/services/SimpleKeyboardIME.kt
```diff
@@ -62,6 +61,9 @@ abstract class SimpleKeyboardIME(
     private var emojiBtnTablet2: Button? = null
     private var emojiSpaceTablet2: View? = null
     private var emojiBtnTablet3: Button? = null
+
+    // how quickly do we have to doubletap shift to enable permanent caps lock
```
andrewtavis, line 65: For comments that are their own line, @Linfye, can we capitalize the first letter and put a period at the end as they should be complete sentences? In the future :) I'll fix this!

andrewtavis: And also you were copying it over, so no stress, just for next time 😊

andrewtavis (approved):
Updated the comments in the project to also follow the style discussed above. Thanks for the fixes here, @Linfye!

andrewtavis:
We're really making some great progress 😊

Issue #231, opened 2024-10-29
Language Settings Switches Doesn't Work


Linfye:
### Terms

- [X] I have searched all [open bug reports](https://github.com/scribe-org/Scribe-Android/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow Scribe-Android's [Code of Conduct](https://github.com/scribe-org/Scribe-Android/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

Currently, the Settings button cannot control the toggling of the Emoji Autosuggest and Double Periods features. I'm not sure when these bugs appeared; I guess it may have been during the previous simplification and refactoring process.

### Device type

Medium Phone API 35 in Android Studio Emulator

### Versions

The latest main branch

andrewtavis:
Closed by #235 🚀 Really appreciate the issue and the work to clear it up, @Linfye!

PR #235, opened 2024-10-31
Fix language settings switches

Linfye:
### Contributor checklist

---

### Description
This PR fixes two bugs: one is the autosuggest switch not working, and the other is the lag when double tap space with `double space periods` switch turned off.

But the fix for the latter one is a bit of a workaround; I’d be happy to revise it if there is a better way.

### Related issue

-   #231


andrewtavis (approved):
Thanks for this, @Linfye! Really appreciate you deducing the problem and coming up with the solution :)

Issue #239, opened 2024-11-09
Move installation corner to the left when menu language is RTL


andrewtavis:
### Terms

- [X] I have searched [open and closed design issues](https://github.com/scribe-org/Scribe-Android/issues?q=is%3Aissue+label%3Adesign)
- [X] I agree to follow Scribe-Android's [Code of Conduct](https://github.com/scribe-org/Scribe-Android/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue would add in an optional top left corner triangle for the installation screen of the menu for RTL languages. As Arabic is the only RTL language in Scribe-i18n at this time, the menu should check to see if Arabic has been selected and then switch the corner triangle if so. The current interface can be seen here:

![Installation Directions](https://github.com/user-attachments/assets/2683cace-986e-4eaf-90e8-7b90ab0c0226)

### Contribution

Happy to support with this and review when a PR is up! Can also work on it myself if needed 😊

Linfye:
I'd like to work on this. 😊

andrewtavis:
Nice, @Linfye! Feel free to use the [update_i18n_keys.sh](https://github.com/scribe-org/Scribe-Android/blob/main/update_i18n_keys.sh) script to bring over the latest version of Scribe-i18n into the project 😊

andrewtavis:
Leaving this open for now until we bring in the localizations, @Linfye :)

andrewtavis:
@Linfye, we now have the Arabic strings in the repo. Can you pull them down and check that the changes in #241 are working? :)

Linfye:
> @Linfye, we now have the Arabic strings in the repo. Can you pull them down and check that the changes in #241 are working? :)

 OK, I'll do this later today. 😊

Linfye:
@andrewtavis  It works fine for me. :)
![PixPin_2024-11-17_21-46-18](https://github.com/user-attachments/assets/b54f73b7-68e9-4f2a-b4c5-e9930a7fca8d)
And do we need to mirror the right_arrow icon in RTL layout? This icon displays on sevral pages.

andrewtavis:
Nice, @Linfye! Ya that change would be great :) Can you send along a PR for it?

Linfye:
> Nice, @Linfye! Ya that change would be great :) Can you send along a PR for it?

Yes, I'd love to after solving #160. 😊

andrewtavis:
[...]

PR #241, opened 2024-11-11
Top left corner triangle for Arabic

Linfye:
### Contributor checklist

---

### Description
When the language is set to Arabic, the "installation corner" on the main page will move to the top-left corner. Since internationalization (i18n) has not yet introduced Arabic, you can use "de" to test the code. 😊

### Related issue

-   #239

### DIFF
--- a/app/src/main/java/be/scri/fragments/MainFragment.kt
+++ b/app/src/main/java/be/scri/fragments/MainFragment.kt
@@ -51,8 +51,13 @@ class MainFragment : ScribeFragment("Main") {
         val cornerCogImageView: ImageView = binding.cornerCog
         val paramsTriangleLayout = binding.cornerTriangle.layoutParams as FrameLayout.LayoutParams
         val paramsCogLayout = binding.cornerCog.layoutParams as FrameLayout.LayoutParams
-        val currentLocale = requireActivity().resources.configuration.locales.get(0)
-        val languageCode = currentLocale.language
+        val currentLocale =
+            requireActivity()
+                .resources.configuration.locales
+                .get(0)
+        val languageCode =
+            currentLocale
+                .language
         if (languageCode == "ar") {
             paramsTriangleLayout.gravity = Gravity.TOP or Gravity.START
             paramsCogLayout.gravity = Gravity.TOP or Gravity.START
--- a/app/src/main/java/be/scri/fragments/MainFragment.kt
+++ b/app/src/main/java/be/scri/fragments/MainFragment.kt
@@ -6,9 +6,12 @@ import android.content.res.Configuration
 import android.os.Bundle
 import android.provider.Settings
 import android.util.Log
+import android.view.Gravity
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
+import android.widget.FrameLayout
+import android.widget.ImageView
 import androidx.activity.addCallback
 import androidx.appcompat.app.AppCompatDelegate
 import be.scri.R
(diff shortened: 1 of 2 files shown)


MahmoudMabrok:
@Linfye  thanks for PR, could you please add screenshot from app before and after the fix.

Linfye:
> @Linfye thanks for PR, could you please add screenshot from app before and after the fix.

Done. :)

app/src/main/java/be/scri/fragments/MainFragment.kt
```diff
@@ -39,9 +42,34 @@ class MainFragment : ScribeFragment("Main") {
...
+            paramsTriangleLayout.gravity = Gravity.TOP or Gravity.START
+            paramsCogLayout.gravity = Gravity.TOP or Gravity.START
+            cornerTriangleImageView.scaleX = -1f
+            cornerCogImageView.scaleX = -1f
+            Log.d("Debug", "Arabic")
+        } else {
+            Log.d("Debug", "Not Arabic")
+        }
+        return languageCode == "ar"
+    }
```
MahmoudMabrok, line 71: we need to make a single responsibility to this method, i.e. making it do single task. 

from it's name, 'isArabicLanguage' , which mean it check the language so it should not have logic related to change the UI configuration such as scale/gravity. 

so I think we would split it into two method, one to check language and other to change the view data. 

checking language would be in separate util function and would be unit tested.

app/src/main/java/be/scri/fragments/MainFragment.kt
```diff
@@ -39,9 +42,34 @@ class MainFragment : ScribeFragment("Main") {
             }
         (requireActivity() as MainActivity).setActionBarButtonVisibility(false)
         callback.isEnabled = true
+        isArabicLanguage()
         return binding.root
     }
 
+    private fun isArabicLanguage(): Boolean {
+        val cornerTriangleImageView: ImageView = binding.cornerTriangle
```
MahmoudMabrok, line 50: what about checking layout direction, so we would cover all RTL languages without need to explicitly check each language.

app/src/main/java/be/scri/fragments/MainFragment.kt
```diff
@@ -39,9 +42,34 @@ class MainFragment : ScribeFragment("Main") {
...
+        val paramsCogLayout = binding.cornerCog.layoutParams as FrameLayout.LayoutParams
+        val currentLocale =
+            requireActivity()
+                .resources.configuration.locales
+                .get(0)
+        val languageCode =
+            currentLocale
+                .language
+        if (languageCode == "ar") {
+            paramsTriangleLayout.gravity = Gravity.TOP or Gravity.START
```
MahmoudMabrok, line 62: for gravity, I think we do not need to handle it, as it already handled in xml, it set to top|end so in RTL it will be automatically be on left of screen.

app/src/main/java/be/scri/fragments/MainFragment.kt
andrewtavis: If this is possible, then yes, but we can also just save a local variable for RTL languages that we support as as of now the plan is just Arabic and eventually Hebrew.

app/src/main/java/be/scri/fragments/MainFragment.kt
andrewtavis: `isArabicLanguage` should also be `isRTLLanguage` so that we can have it be general from the start :)

Linfye:
@MahmoudMabrok Could you review it again? I think this is what you mentioned before. 😊

MahmoudMabrok:
for me all is fine, just for scalability ,  if we can use `if (Locale.getDefault().layoutDirection == LayoutDirection.RTL)` instead of hardcore the names of languages. 

if not stable solution, so PR is fine for me.

andrewtavis:
This makes sense to me as well, as we actually have `de` in there now on accident, so let's go with RTL. Thanks for the suggestion, @MahmoudMabrok! @Linfye, do you want to make the change and then I can merge it in?

MahmoudMabrok (approved):

MahmoudMabrok:
> > This makes sense to me as well, as we actually have `de` in there now on accident, so let's go with RTL. Thanks for the suggestion, @MahmoudMabrok! @Linfye, do you want to make the change and then I can merge it in?
> 
> Is this OK now? 😊
> 
> cc @MahmoudMabrok

Thanks so much for your efforts. You did a great work. 

For me, it is ready now.

andrewtavis (approved):
Awesome, thanks @Linfye and @MahmoudMabrok! Let's give this another check post localizations and bringing in the Arabic texts 🌐

Issue #243, opened 2024-11-16
building app failing duo to missing string resource 


MahmoudMabrok:
### Terms

- [X] I have searched all [open bug reports](https://github.com/scribe-org/Scribe-Android/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow Scribe-Android's [Code of Conduct](https://github.com/scribe-org/Scribe-Android/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

pipeline failed to build app duo to missing string resources. 

https://github.com/MahmoudMabrok/Scribe-Android/actions/runs/11871381606/job/33083823181#step:4:78

### Device type

NA

### Versions

Latest

andrewtavis:
Thanks for the issue, @MahmoudMabrok! Definitely unexpected given Scribe-i18n is at 100% across all languages in the most recent import. CC @angrezichatterbox :)

MahmoudMabrok:
it was building app before normally, but now it keeps failing.

MahmoudMabrok:
I tried tp do dummy pr and check current workflow and it fails also. 
check [here](https://github.com/MahmoudMabrok/Scribe-Android/actions/runs/11871440501/job/33083955591#step:4:77) and [here](https://github.com/MahmoudMabrok/Scribe-Android/actions/runs/11871440498/job/33083955589?pr=2#step:5:137)

angrezichatterbox:
> Thanks for the issue, @MahmoudMabrok! Definitely unexpected given Scribe-i18n is at 100% across all languages in the most recent import. CC @angrezichatterbox :)

For now, the Gradle task does not move from the i18n folder to the res. I will look into it and make a quick fix by tonight.

angrezichatterbox:
The issue must have occurred during the time I made the conversion from build.gralde to build.gradle.kts. Thanks @MahmoudMabrok for finding this issue.

MahmoudMabrok:
the issue is [here](https://github.com/scribe-org/Scribe-Android/commit/8c70a6a9dc812dbe339c3b855cd3c8025c262bf9#diff-1c5d103c04afd91ef1dc29f75921c46985eeafcedf2b88b7bfedf61a0914db48L52) this string  is removed.

PR #245, opened 2024-11-16
fix:The Build Task for i18n to res folder not working properly

angrezichatterbox:
### Contributor checklist

---

### Description

The gradle task for transferring files from i18n to the res folder for localization was not transferring files, which caused the i18n update to not be reflected in the res folder. 

### Related issue

-   #243


MahmoudMabrok (approved):

app/build.gradle.kts
```diff
@@ -201,18 +201,23 @@ dependencies {
 }
 
 tasks.register<Copy>("moveFromi18n") {
-    val locales = listOf("de", "es", "sv", "en-US")
+    duplicatesStrategy = DuplicatesStrategy.INCLUDE
+    destinationDir = file("src/main/res")
 
+    val locales = listOf("de", "es", "sv", "en-US")
```
andrewtavis, line 207: Do we need to expand this with the other localizations that are included in Scribe-i18n?

MahmoudMabrok: val  locales = file("src/main/assets/i18n/Scribe-i18n/values").listFiles()
        .findAll { it.isDirectory() }
        .collect { it.name }
        
could we use something like this

andrewtavis:
Appreciate the help, @MahmoudMabrok! Your version of getting the locales worked great, and from there we needed to add them all to the UI. This has been done by hand so far, but maybe we can figure out a way to load those dynamically as well 😊

andrewtavis (approved):
Thanks for the quick fix here, @angrezichatterbox, and also for the collaboration to get it all working, @MahmoudMabrok! 😊

Issue #160, opened 2024-09-30
Switch enter button icon to play button when in a command mode


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Android/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-Android's [Code of Conduct](https://github.com/scribe-org/Scribe-Android/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

One of the usability features of Scribe is that the enter key switches color and its icon when the user is in a command mode to enter a command. As of now when a user presses the Scribe key and then a command button they're shown the command bar and they'll then type in the command bar as well. The enter key is then Scribe blue, but we also want that its icon switches to a play button. The play button icon should be filled in - i.e. not the just the boarder. An example of the interface in Scribe-iOS can be seen in the Figma designs [here](https://www.figma.com/design/c8945w2iyoPYVhsqW7vRn6/scribe_public_designs?node-id=403-9450&t=QznVUA7nMwE5CZge-1) (the interface over translation) and below:

![iPhone 14 White Minimal](https://github.com/user-attachments/assets/912e4f2c-86e0-4b4b-908f-b9708c3eff0c)

### Contribution

This is a great good first issue! Happy to review or support someone wanting to get into Scribe-Android development as needed 😊

KekmaTime:
@andrewtavis hey i would love to work on this issue as part of hacktoberfest

andrewtavis:
Sure thing, @KekmaTime! Let us know if there's anything we can do to support 😊

andrewtavis:
Hey @KekmaTime 👋 Please let us know if you need any support here :)

Linfye:
@KekmaTime Do you still want to work on this issue? Can I participate or collaborate? 😊

cc @andrewtavis

andrewtavis:
Thanks for passing it off, @KekmaTime, and thanks for picking it up, @Linfye! 🚀

andrewtavis:
Wouldn't it be setting the key icon and then reloading the keyboard whenever the keyboard is in a command mode? What have you tried so far, @Linfye?

Linfye:
I tried to add a judgment in the keyboard build helper, and then dynamically update the value of isCommandMode in the SimpleKeyboardIME, but the keyboard did not update the icon after this value was updated.

```
if (key.code == KEYCODE_ENTER) {
[...]

PR #251, opened 2024-11-21
Play Icon When in a Command Mode

Linfye:
### Contributor checklist

---

### Description
This PR changes enter icon to play button when in a command mode.

### Related issue

-   #160


Linfye:
The creation process of the keyboard is implemented by a helper, and I haven't found a suitable way to pass the `isCommandMode` parameter. After modifying the `isCommandMode` value in SimpleKeyboardIME and recreating the keyboard, it seems that the parameter is reset, and I don't know how to solve this. Could you help me with this? Thanks! :) @andrewtavis

andrewtavis:
CC also @angrezichatterbox as I'm at work atm :) I'll take a look tonight or over the weekend, @Linfye! Thank you!

angrezichatterbox:
I am not sure we have any easy ways of passing the command boolean. However, I have another way in which we could do it. One method is to set a custom state within the keyboard. So for eg if we use the  EditorInfo.IME_ACTION_UNSPECIFIED is the condition for setting the icon of the enter key to the Play Icon. Now there is the setEnterKeyColor function use it to set the mEnterKeyType as that specific type. Now create a function to set the enter key icon using this logic in MyKeyboard.kt and call this function after setting the mEnterKey type. You can use any custom type to do it.

I am happy to explain in more detail :)

Linfye:
> I am not sure we have any easy ways of passing the command boolean. However, I have another way in which we could do it. One method is to set a custom state within the keyboard. So for eg if we use the EditorInfo.IME_ACTION_UNSPECIFIED is the condition for setting the icon of the enter key to the Play Icon. Now there is the setEnterKeyColor function use it to set the mEnterKeyType as that specific type. Now create a function to set the enter key icon using this logic in MyKeyboard.kt and call this function after setting the mEnterKey type. You can use any custom type to do it.


I had considered this method initially, but I thought that modifying the built-in state might not be a common practice in Android development (I'm not quite sure about this). Since you suggested it, I can give it a try. Thank you! :)

angrezichatterbox:
> > I am not sure we have any easy ways of passing the command boolean. However, I have another way in which we could do it. One method is to set a custom state within the keyboard. So for eg if we use the EditorInfo.IME_ACTION_UNSPECIFIED is the condition for setting the icon of the enter key to the Play Icon. Now there is the setEnterKeyColor function use it to set the mEnterKeyType as that specific type. Now create a function to set the enter key icon using this logic in MyKeyboard.kt and call this function after setting the mEnterKey type. You can use any custom type to do it.
> 
> I had considered this method initially, but I thought that modifying the built-in state might not be a common practice in Android development (I'm not quite sure about this). Since you suggested it, I can give it a try. Thank you! :)

You could also set a companion object as a new type and set the key type as that from the setEnterKeyColor function.

Linfye:
@angrezichatterbox Sorry for the delay, I've been busy with school matters lately. And thank you very much for your help! 

Could you review the current code? 😊

andrewtavis:
Functionality wise it looks good to me, @Linfye :) Might be nice if we could have the play icon moved a bit to the right - say 10% of the button's width, as with these sorts of icons if we actually center it it doesn't "look centered" :) This could be another task if you'd like it to be though!

Linfye:
> Functionality wise it looks good to me, @Linfye :) Might be nice if we could have the play icon moved a bit to the right - say 10% of the button's width, as with these sorts of icons if we actually center it it doesn't "look centered" :) This could be another task if you'd like it to be though!

Yes, I noticed this issue before but forgot to mention it here. I can update this icon by adding a transparent margin on the left side to resolve it. What do you think? :)

andrewtavis:
That'd be perfect, @Linfye! Thank you!

angrezichatterbox (commented):
@Linfye Everything looks nice. We could make the visual changes as suggested. We also have an edge case to deal with which comes which the approach. This could be an issue which could be created and worked on later on. 

So currently when the user closes the keyboard when in any of the modes where the play icon is shown. If the user switches to an app and gets back to keyboard the play icon disappears and it would become again the icon as per the Editor Info. 

A thing we could do is to switch the mode to idle when the user exits off the keyboard. Would love to hear about this specific issue. 

Other than this amazing work @Linfye

andrewtavis:
Switching to idle when the user navigates away makes sense to me, @angrezichatterbox 😊

angrezichatterbox:
> Switching to idle when the user navigates away makes sense to me, @angrezichatterbox 😊

Yes it makes sense. Should we create an issue for correcting this edge case after this PR is merged ?

andrewtavis:
Either or. @Linfye, if you want to include it in here, then by all means, or if not let's make the issue 😊

angrezichatterbox:
@Linfye Would u like to bring it in or should I add it to this PR :) Happy to add it to the PR :)

Linfye:
> @Linfye Would u like to bring it in or should I add it to this PR :) Happy to add it to the PR :)

Yes, I'd like to solve this.

Linfye:
@angrezichatterbox I've tried to change the keyboard state using `onFinishInput`, but I'm not sure how to update the UI. There's some strange functionality disorder. Would you like to take over this?

angrezichatterbox:
> @angrezichatterbox I've tried to change the keyboard state using `onFinishInput`, but I'm not sure how to update the UI. There's some strange functionality disorder. Would you like to take over this?

I would look into it and would get back to you.

angrezichatterbox:
So the function that was being run when the input view  is closed was onFinishInputView(). I have implemented the changes. I just checked the codebase a bit more now. Thanks for replacing the deprecated setBackgroundDrawable(). 

It is ready for review @andrewtavis

Issue #247, opened 2024-11-17
Migrate the settings page to Jetpack Compose 


angrezichatterbox:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Android/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-Android's [Code of Conduct](https://github.com/scribe-org/Scribe-Android/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

The current Settings page is implemented using XML and a Fragment. This issue involves migrating the Settings page to Jetpack Compose. The goal is to replace the XML layout with a Compose-based UI and update the fragment's content view to use Compose. This ensures that the app remains functional throughout the migration, and `ViewPager2` continues to work during the process.

The migration will replace `RecyclerView` with Compose alternatives like `LazyRow` and `LazyColumn`, removing the need for the existing adapter. For more details on handling `LazyRow` and `LazyColumn`, refer to [this document](https://developer.android.com/develop/ui/compose/lists).

### Contribution

Happy to support in solving this issue or take this up :)

angrezichatterbox:
I will be working on this issue. @Jag-Marcel I would solve this issue and make a PR in a day or two. It would give some more idea on the exact change to be done for the conversion of XML to Compose and what exactly needs to be done for #248

Saifuddin53:
Hello, @angrezichatterbox if you haven't yet started this migration then can I work on it?

angrezichatterbox:
I have started working on this issue for now and would make a PR in a day or two. Would you like to check out some of the other issues.

Saifuddin53:
Sure! I'm definitely interested @angrezichatterbox, but I'd be especially keen to work on any issues related to migrating to Jetpack Compose if there are any.

Saifuddin53:
@angrezichatterbox can I create an issue for migration of privacy policy screen and then start working on it?

angrezichatterbox:
The page is already being worked on at the moment. Depending on the requirement the language settings page might need a separate issue but I'm not sure for now as it might be completed within this issue if we could reuse the existing created components. I'll be creating the issue for the language settings page if it's needed in the next few days.

Saifuddin53:
Okay no problem I would like to work on any other issue as well

angrezichatterbox:
Would you like me to suggest you some issues to checkout :)

Saifuddin53:
[...]

PR #252, opened 2024-11-27
Migrate the settings page to Jetpack Compose

angrezichatterbox:
### Contributor checklist

---

### Description

The pull request moves the settings page to Jetpack Compose. A separate issue would be created to migrate the Language settings page to Jetpack Compose since it’s too large to fit within the same PR. 

### Related issue

#247


angrezichatterbox:
Happy to make changes to any visual detail that I might have missed while migrating to Jetpack Compose.

andrewtavis:
Generally all's looking really good, @angrezichatterbox :) One thing I'm seeing is that the divider line for language sub menus isn't quite right? Should be more in the middle? See below:

Screenshot 2024-11-27 at 19 56 56

Aside from that, would you suggest that the toggles change style as they have? The new versions have a cta orange outline if they're not active rather than just a grey background, and there's also padding around the circular interactive element of the toggle, where before it there was no padding. Happy to hear your feedback on this and we can go from there! Would just be good to make sure that iOS and Android are the same, but we can work on it :)

angrezichatterbox:
> Generally all's looking really good, @angrezichatterbox :) One thing I'm seeing is that the divider line for language sub menus isn't quite right? Should be more in the middle? See below:
> 
> Screenshot 2024-11-27 at 19 56 56
> 
> Aside from that, would you suggest that the toggles change style as they have? The new versions have a cta orange outline if they're not active rather than just a grey background, and there's also padding around the circular interactive element of the toggle, where before it there was no padding. Happy to hear your feedback on this and we can go from there! Would just be good to make sure that iOS and Android are the same, but we can work on it :)

So currently I am directly using Material Switches I would make it to use the earlier switch. I missed it during the migration. I would also make the divider towards the centre. Thanks :)

andrewtavis (approved):
Thanks so much for the word here, @angrezichatterbox! In looking at [the toggle designs I made](https://www.figma.com/design/c8945w2iyoPYVhsqW7vRn6/scribe_public_designs?node-id=513-2072&t=IC3mMCgbuXzrmjUe-1) the new versions are much better than the old ones 😊 Thanks for the attention to detail here and continuing the migration!

Issue #214, opened 2024-10-20
Add in all needed interface variable strings to the keyboard constants


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Android/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-Android's [Code of Conduct](https://github.com/scribe-org/Scribe-Android/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

Post a recent discussion in a Scribe dev sync, we decided that all variables that are required for functioning keyboards should be hard coded into the application rather than localized via Weblate. Based on this, we need the following strings hard coded into each keyboard file, with the examples for Spanish being below:

- currencySymbol = "$"
- spaceBar = "espacio"
- language = "Español"
- invalidCommandMsg = "No en Wikidata"
- translateKeyLbl = "Traducir"
- translatePlaceholder = "Ingrese una palabra"
- translatePrompt = commandPromptSpacing + "es -› \(getControllerLanguageAbbr()): "
- translatePromptAndCursor = translatePrompt + commandCursor
- translatePromptAndPlaceholder = translatePromptAndCursor + " " + translatePlaceholder
- + the equivalents for this for conjugate and plural as well

See the [Scribe-iOS interface variables for Spanish](https://github.com/scribe-org/Scribe-iOS/blob/2d60078480cdae56e6cc2eac0375fcb1c62e0e09/Keyboards/LanguageKeyboards/Spanish/ESInterfaceVariables.swift#L258) for further information :)

### Contribution

Happy to discuss the needs here and review when a PR is up 😊

Saifuddin53:
Great then I would like to get started with this issue @angrezichatterbox

andrewtavis:
Thanks for you vested interest in working with us, @Saifuddin53! Please let us know if you have any questions 😊

Saifuddin53:
Starting this one @andrewtavis but before that it would be great if you could provide me with some more context about how we are going to use these constant variables as I am not able to use the ios version of this app I couldn't get too much from the link you provided 🙂

andrewtavis:
Those variables are basically going to be used within the interface of the keyboard, @Saifuddin53 :) So some explanations of the above are:

```
  currencySymbol = // base currency symbol for the keyboard
  currencySymbolAlternates = // which alternate characters should the above key have when we long press it
  spaceBar = // what's space called in the language
  language = // what's the name of the language
[...]

PR #255, opened 2024-12-01
Added interface variables strings constant to be used in the keyboard 

Saifuddin53:
### Contributor checklist

---

### Description

In this PR, I’ve added all required interface variable strings to the keyboard constants for each language. Based on the new approach, these strings are now hard-coded into each keyboard file rather than being localized via Weblate.

### Related issue

-   #214

### DIFF
--- a/app/src/main/java/be/scri/helpers/english/ENInterfaceVariables.kt
+++ b/app/src/main/java/be/scri/helpers/english/ENInterfaceVariables.kt
@@ -1,7 +1,6 @@
 package be.scri.helpers.english
 
 object EnglishLanguageConstants {
-
     // Currency Symbol and Alternates
     const val currencySymbol = "$"
     val currencySymbolAlternates = listOf("$", "€", "£", "¥", "₩", "¢")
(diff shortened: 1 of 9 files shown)


Saifuddin53:
@andrewtavis @angrezichatterbox I've implemented all the interface variables for Spanish (for now). Could you please review this and let me know if this is the implementation you had in mind? Please let me know if any modifications are needed or if you have any suggestions 🙂

angrezichatterbox:
Looks good to me. The organization is perfect for constants.

Saifuddin53:
@angrezichatterbox @andrewtavis ready for review

andrewtavis:
Could we discuss really quick the [linting errors](https://github.com/scribe-org/Scribe-Android/actions/runs/12107001558/job/33753396129?pr=255) that are suggesting that we do screaming snake case for variables that won't change? That makes sense to me :) I guess that Swift doesn't have that convention, but if that's what ktlint suggests then I'd say we'd be good to follow it 😊

Saifuddin53:
@andrewtavis I’ve updated the variable names to follow the screaming snake case convention. Thank you for your feedback ☺️

andrewtavis:
What I get for using VS Code to do something quick 😅 I'll fix the linting errors and then bring this in :)

angrezichatterbox:
If we are migrating the keyboard to compose we would have to rework the entire MyKeyboardView.kt and The IME. I am not sure if it is something necessary to be done as most keyboards are still using XML layouts as it wouldn't be easy to work with the IME using Compose.

Issue #269, opened 2024-12-15
[Conjugation] Possible reference of conjugation/tense labels via a conjugation metadata JSON file


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Android/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-Android's [Code of Conduct](https://github.com/scribe-org/Scribe-Android/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

A question that's coming to mind right now is how Scribe-Android (and Scribe-iOS) will deal with new conjugations coming in from Scribe-Data via Scribe-Server. Scribe-Data will be getting new data as it's added to Wikidata, and this could include conjugations that have yet to be added into a Scribe keyboard app. This data would then be transferred to a Scribe keyboard application (and Scribe Conjugate applications, for that matter), but then the needed labels for what the conjugation is and which conjugation view should be used would then be unknown by the interface. This would mean that Scribe-Data would be providing new data, but the end applications would need a version update in order to reference it.

One way of handling this would be to include JSON file in the Scribe-Server response that as of now could be hard coded. This metadata file would be loaded into the application and used to define conjugation tenses:
- What's the label of the conjugation tense - i.e. "present simple", "past perfect", etc
- Which fields returned in a verbs response data are included in this tense
- The UI would then be designed in a way that we would not need to hard code the conjugation logic anymore, but rather it would be loaded from the metadata file and the conjugation view would be determined by the number of conjugations in the tense

### Contribution

Happy to discuss this and get feedback on the idea! Once we've decided on the functionality, we can then work to add this file and have the work in #268 be based on it :)

andrewtavis:
@angrezichatterbox: Would be great to get your opinion here :) Let me know if the above isn't clear! Just realized that we're going to have this problem if the data expands and then we don't have a way of remotely updating the fields.

andrewtavis:
CC @wkyoshida, @henrikth93, @mhmohona and @Jag-Marcel as well, as this is a major infrastructure discussion that would affect all Scribe keyboard and conjugate apps 😊

angrezichatterbox:
I am not sure if what I have understood is right or wrong. But we would be constructing the conjugation view based on the JSON response we would be getting from Scribe server. Am I right ?

henrikth93:
[...]

PR #293, opened 2025-01-05
#269 First iteration of Scribe-Data data contracts

andrewtavis:
### Contributor checklist

---

### Description

As discussed in #269 and in recent syncs, we need to set up data contracts that will determine how the end applications present the data coming in from Scribe-Data. This PR sends along my first ideas on them based on the call we had yesterday.

A short synopsis of this:

- For Scribe-iOS all data values are hard coded into the application
- The data might change on Wikidata's end, meaning that the labels and other information would change in Scribe-Data
    - Ex: More data being added, which as of now would require us to create a new app release to change the hard coded fields meaning that people would not have access to the new data till then
    - Ex: As of now - unlike other languages - Spanish has the feminine and masculine of nouns stored on the same lexeme
        - If this changes, then we'd then have to rework the entire way that data is accessed and the Spanish keyboard would break on the next data update requiring an update of the app

Discussed solution:
- We want to make sure that data updates are completely separate from app updates
- We create "data contracts" that are delivered with new data packs
    - Ex: User downloads Scribe-iOS/Android and wants the German keyboard
        - They get the German data and the German contract
        - Updates of the data would deliver the corresponding contract
- Note that we'd need to set up tests for Scribe-Data whereby the data that should go into the contract is changed
    - Ex again for Spanish nouns
        - Say they're split and we don't have feminine and masculine on the same lexemes
        - We check the data outputs and see if the data that's required to fulfill the current contract is coming in
        - If not, we open an issue that says what parts of the contract aren't being fulfill and we then check to see how the data has changed

Feedback is very welcome on the above! 🙏

### Related issue

-   #269

### DIFF
--- a/app/src/main/assets/data-contracts/de.json
+++ b/app/src/main/assets/data-contracts/de.json
@@ -1,0 +1,39 @@
+{
+    "numbers": { "nominativeSingular": "nominativePlural" },
+    "genders": {
+        "canonical": ["gender"],
+        "feminines": [],
+        "masculines": [],
+        "commons": [],
+        "neuters": []
+    },
+    "conjugations": {
+        "1": {
+            "title": "Präsens",
+            "1": { "ich": "indicativePresentFirstPersonSingular" },
+            "2": { "du": "indicativePresentSecondPersonSingular" },
+            "3": { "er/sie/es": "indicativePresentThirdPersonSingular" },
+            "4": { "wir": "indicativePresentFirstPersonPlural" },
+            "5": { "ihr": "indicativePresentSecondPersonPlural" },
+            "6": { "sie/Sie": "indicativePresentThirdPersonPlural" }
+        },
+        "2": {
+            "title": "Preterite",
+            "1": { "ich": "indicativePreteriteFirstPersonSingular" },
+            "2": { "du": "indicativePreteriteFirstPersonPlural" },
+            "3": { "er/sie/es": "indicativePreteriteSecondPersonSingular" },
+            "4": { "wir": "indicativePreteriteSecondPersonPlural" },
+            "5": { "ihr": "indicativePreteriteThirdPersonSingular" },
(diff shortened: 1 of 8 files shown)


andrewtavis:
Again the goal of this is that we get rid of as much of the "command variables" files from Scribe-iOS that are hard coded.

Noun gender annotation:
- Is there a canonical gender column (there is for all languages except Spanish as of now)
- Check the singular column for the word and return it's gender
- Also check if it's in the plural column
- For Spanish we need to check to see if it's in the feminine or masculine singular columns given the contract

Plural command:
- Check if word is in the columns that are the value(s) of `numbers` for "Already plural"
- If not, check the key columns and return the respective value column

Conjugate command:
- The conjugations are presented to the user in the order of the dictionary object
- The title is what appears as the header in the UI
- The keys for the conjugation sub objects are the labels of the conjugation field
- The values are the value that's returned when the user presses the conjugation key
    - Return value from this column

Translate command:
- Is ultimately the easiest as the plan is that we have a word and the data that we have for it is the contract to be fulfilled
- Is the word just a noun, a verb, another type or more than one?
- Then what are the options and their descriptions
    - Ex German translation: Book -> noun or verb? -> noun -> Buch (book) or Roman (novel)

CC @axif0 as well as this effects Scribe-Data work to come :)

andrewtavis:
@henrikth93 and @Jag-Marcel: You having worked on iOS the most, basically mapping out files like [Scribe-iOS/Keyboards/LanguageKeyboards/English/ENCommandVariables.swift](https://github.com/scribe-org/Scribe-iOS/blob/main/Keyboards/LanguageKeyboards/English/ENCommandVariables.swift). Not all of those files will be put into the contract, but everything that can change with the data :)

angrezichatterbox:
I feel the formatting is quite good for now. As far as the experimentation I have with finding the gender of a word. I think this contract makes it easier and proper column would be used for the word depending on the data available.

andrewtavis:
Merging this in for now 😊

Issue #272, opened 2024-12-15
[Plural] Insert noun plural when a user enters a noun into plural command and hits enter


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Android/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-Android's [Code of Conduct](https://github.com/scribe-org/Scribe-Android/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue covers the plural command functionality of the Scribe-Android keyboards. 

- Note that this issue is blocked by #271 where the plural text proxy label and placeholder text will be added

The functionality would be:

- The user presses the Scribe key on a keyboard
- The user selects the plural key option and is presented with the label and placeholder from the work in #271
- The user enters a noun that is contained in the nouns table in the language database for the given keyboard
- The user hits the enter key (return key) and the plural for the noun is inserted into the text proxy

### Contribution

Happy to discuss this and review functionality once a PR is up! 🚀

smalik2811:
Hi, since the issue #271 has been closed, I would like to work this issue.

andrewtavis:
We need to merge in the current PRs for this to be unblocked, but I'll get to those today/tomorrow and then we can work on this :) Thanks, @smalik2811! I'll assign and unblock, but again let's wait till both your and @angrezichatterbox's PRs are in :)

andrewtavis:
@angrezichatterbox's PR adds in the way that we'll be using the contracts to get data from the tables, which you can follow for this issue 😊

smalik2811:
I have never heard about the contracts before. I do not know what they mean or what's their purpose, @andrewtavis could you please share a link to some reference material which I can refer to in order to learn more about them 😊.

andrewtavis:
Hey @smalik2811 👋 Data contracts are a very new thing for the community :) I just wrote up a MD file to answer your questions on this :) You can find it at [scribe-org/Organization/DATA_CONTRACTS.md](https://github.com/scribe-org/Organization/blob/main/DATA_CONTRACTS.md). Please let me know if you have questions and I'll update the doc!

smalik2811:
When the user has switched to command mode and clicks on the `space` button, then instead of adding the space to the command bar text field, the spaces are added at the end of the input connection. I have included a video for reference.

[...]

Issue #270, opened 2024-12-15
[Conjugation] Insert a conjugation when a conjugation key is pressed


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Android/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-Android's [Code of Conduct](https://github.com/scribe-org/Scribe-Android/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue is the final issue to finish the conjugation functionality. It's blocked by the following issue:

- #267 
- #268 
- #269 

The work for this issue is inserting a conjugation from a conjugation key into the text proxy.

### Contribution

Happy to support as I can and review once a PR is up! 🚀

Issue #266, opened 2024-12-14
[Translation] Return translation based on an entered value and pressing the return key in Translate Mode


angrezichatterbox:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Android/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-Android's [Code of Conduct](https://github.com/scribe-org/Scribe-Android/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue aims to integrate the Translate feature into the Scribe-Android keyboard. When users type into the keyboard in Translate mode and add text to the command bar, pressing the return key should automatically translate the text from the translation source language to the keyboard language.

Currently, this issue is blocked by:

- #149
- #265 

The above issues propose adding a menu option for the translation source language and switching the keyboard to the translation source language keyboard when in Translate mode. 

### Contribution

Happy to help anyone solve this issue or take this up myself at some point :)

andrewtavis:
@angrezichatterbox: I edited the issue text above to keep the blocked issues as unordered list elements on their own line so that their name and status are rendered in the issue text. This makes it easier to see when the issue is unblocked :)

angrezichatterbox:
I will work on this issue :)

andrewtavis:
Thanks @angrezichatterbox! With this we'll be at two commands! Exciting stuff 😊

PR #299, opened 2025-01-17
Fix(general-keyboard): Automatically switch to letter keys after command selection

### LINKED ISSUE #298: Switch keyboard to letters when a command mode command is selected
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Android/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-Android's [Code of Conduct](https://github.com/scribe-org/Scribe-Android/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

When the user selects a Scribe command it would be good if the keyboard switches to the letter keys as the next thing they would do is enter a word. The workflow would be:

- The user switches to the number/symbol keys
- The user presses the Scribe key and see the translate, conjugate and plural commands
- The user selects a command
- The keyboard mode is switched to letter keys automatically

### Contribution

Happy to assist with this and review when a PR is opened! 😊

smalik2811:
### Contributor checklist

---

### Description

This commit addresses issue #298 by:

1. **Automatically switching the keyboard to 'keyboardLetters' after a Scribe command (e.g., translate, conjugate, plural) is selected.** This improves the user experience by streamlining the workflow and reducing the need for manual keyboard mode changes.

2. **Switching the keyboard to 'keyboardLetters' when the user switches to the Idle mode.** This ensures consistent keyboard behavior and prevents unexpected input modes.

This enhancement enhances the overall usability of Scribe by providing a more intuitive and efficient interaction flow for users.

Closes #298

### Related issue

-   #298

### DIFF
--- a/app/src/main/java/be/scri/services/GeneralKeyboardIME.kt
+++ b/app/src/main/java/be/scri/services/GeneralKeyboardIME.kt
@@ -249,6 +249,7 @@ abstract class GeneralKeyboardIME(
         keyboardView = keyboardBinding.keyboardView
         keyboardView!!.setKeyboard(keyboard!!)
         keyboardView!!.mOnKeyboardActionListener = this
+        handleModeChange(keyboardSymbols, keyboardView, this)
         keyboardBinding.scribeKey.setOnClickListener {
             currentState = ScribeState.IDLE
             switchToCommandToolBar()
@@ -259,6 +260,7 @@ abstract class GeneralKeyboardIME(
     }
 
     private fun setupIdleView() {
+        handleModeChange(keyboardSymbols, keyboardView, this)
         val sharedPref = getSharedPreferences("app_preferences", Context.MODE_PRIVATE)
         val currentNightMode = resources.configuration.uiMode and Configuration.UI_MODE_NIGHT_MASK
         val isSystemDarkMode = currentNightMode == Configuration.UI_MODE_NIGHT_YES
--- a/app/src/main/java/be/scri/services/GeneralKeyboardIME.kt
+++ b/app/src/main/java/be/scri/services/GeneralKeyboardIME.kt
@@ -246,10 +246,10 @@ abstract class GeneralKeyboardIME(
                 keyboardBinding.topKeyboardDivider.setBackgroundColor(getColor(R.color.special_key_light))
             }
         }
+        handleModeChange(keyboardSymbols, keyboardView, this)
         keyboardView = keyboardBinding.keyboardView
         keyboardView!!.setKeyboard(keyboard!!)
         keyboardView!!.mOnKeyboardActionListener = this
-        handleModeChange(keyboardSymbols, keyboardView, this)
         keyboardBinding.scribeKey.setOnClickListener {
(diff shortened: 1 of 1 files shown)


andrewtavis:
Minor feedback here, @smalik2811: I think that we don't need the keyboard switching back to the letter keys each time the state goes to idle. Wasn't quite as clear on this as I could have been :)

An example would be:

- The user is on symbols
- The user presses the Scribe key
- The user presses exit as they're exploring the keyboard for the first time or just decided not to use a command
- In this case the keyboard should still be symbols

What I meant was that each time after the user executes a command that we should switch back to the letter keys, but then that would be handled in the functions that execute the commands, not on a switch to letters whenever the keyboard switches to idle mode.

Do you want to send along the change such that this just switches the keyboard to letters when we enter a command mode, and then we can keep the letter keys switch in mind for when we work on the command functions in #266, #270 and #272?

Thanks for the help with this! 😊

CC @angrezichatterbox on this as well 👋

angrezichatterbox:
So does this mean that the keyboard should be on Keys on all modes except select command mode and idle mode. Like when the user click on a particular command  translate plural or conjugate he is switched to the keys. Am I right ?

andrewtavis:
Yes, exactly, @angrezichatterbox :) If they're on symbols and want to enter a word into a command, then the first thing they'll need to do is switch to letters, and they'll need to do it basically always, so let's do it for them 😊

smalik2811:
Hey @andrewtavis, thanks for the feedback 😅. I totally missed the point that the user can return back to the Idle mode straight from the Command Select mode and this could result in the issue of unnecessaryly switching back from symbols to letters.

I have removed the part causing the issue. Now changes only includes the code to switch symbols to letters then the user actually enters the command mode.

The changes were affecting the capitalization of Noun in German, I have also fixed that.

andrewtavis:
So far so good, @smalik2811! Thanks for the quick reaction :) One final comment here is that the German noun capitalization is still being effected by this, or maybe it's broken and we need to fix it. When we do the plural command on the German keyboard the keyboard should switch to letter keys and it should also be capitalized for one key.

Do you want to try to send along a fix for this? 😊

smalik2811:
Hey @andrewtavis, I tested the app on a virtual device. I first switched to the symbol keyboard from the idle mode and then entered the Plural mode. The keyboard switched to the letters keyboard and the letters are capitalized as expected.


[German capitalization.webm](https://github.com/user-attachments/assets/61ca51cf-c6c6-4000-b789-2e73aad73e28)


I must have missed something 😅, could you specify the issue you noticed with the German noun capitalization.

andrewtavis (approved):
Sorry for the confusion, @smalik2811! Thanks much for the video showing that all's working. I just updated to Android Studio Ladybug and I think that there was just some issues where the changes weren't being reflected. Needed to leave the keyboard and come back to it and it's all great now 😊

Thanks for the continued support!

smalik2811:
No problem at all! Glad you got it sorted out and we have resolved another issue 😊

Issue #135, opened 2024-09-21
[Annotation] Style first autosuggestion button based on the gender of a word typed


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Android/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-Android's [Code of Conduct](https://github.com/scribe-org/Scribe-Android/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue is the first to start adding in Scribe command functionality to Scribe-Android 🚀 

When a user types a word that is found in a nouns table we'd like the first autosuggestion button to change based on the gender of the word. The suggestion key should also be disabled when an annotation is being shown.

The changes are the following:

- For feminine the key is red with an F
- For masculine the key is blue with an M
- For neutral the key is green with an N
- For common (combination of feminine and masculine in Swedish) the key is purple with a C
- For plural the key is orange with a PL

The letters described above should also change based on the language. See [Scribe-iOS/Keyboards/KeyboardsBase/ScribeFunctionality/Annotate.swift#L37](https://github.com/scribe-org/Scribe-iOS/blob/ba8da43a5d3f791a9567cb8c6ff6496c85350987/Keyboards/KeyboardsBase/ScribeFunctionality/Annotate.swift#L37) for the needed conversions for Swedish and Russian.

- Note that this issue is blocked by #134, which is blocked by https://github.com/scribe-org/Scribe-Data/issues/142

### Contribution

Happy to support with this and get to it myself eventually if need be! This is the first Scribe functionality within Scribe-Android. Really looking forward to it 😊

ZayJackPrvy:
Hi @andrewtavis . I'd like to work on this issue. Could you please assign it to me?

andrewtavis:
Hey @ZayJackPrvy 👋 Appreciate your interest in working with us! This issue's currently blocked a few issues in our data ingestion application. You'd be welcome to work on it after, which should be in a few weeks :) Is there another one you'd be interested in in that time?

ZayJackPrvy:
@andrewtavis Yes. I would like to work on an issue that’s not too trivial, such as simple text changes or button adjustments. Could you recommend an appropriate issue for me? It's part of university course to contribute to an open-source project, and I’ve chosen your project. Hopefully it's OK for you.

andrewtavis:
[...]

Issue #161, opened 2024-09-30
Dark mode coloration of command bar


andrewtavis:
### Terms

- [X] I have searched [open and closed design issues](https://github.com/scribe-org/Scribe-Android/issues?q=is%3Aissue+label%3Adesign)
- [X] I agree to follow Scribe-Android's [Code of Conduct](https://github.com/scribe-org/Scribe-Android/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue corrects a few issues with the command UI for the keyboards in dark mode:

1. The Scribe key icon should be white instead of black as shown below

![Screenshot_20240930_182219](https://github.com/user-attachments/assets/d3bd9cd2-4425-43e8-b501-852d7af94261)

2. The `X` to cancel the command selection or a command should also be white instead of black as is shown below
3. The color of the command keys should match the Scribe key and both should be dark Scribe blue (now the command keys are a bit lighter as shown below)

![Screenshot_20240930_182148](https://github.com/user-attachments/assets/c73d701a-c47c-46b3-821a-b40c4beb378a)

4. The command bar should be the same color as the rather than white as shown below
    - The color should be `2E2E2E`, the same color as the special keyboard keys in dark mode like delete, shift, etc

![Screenshot_20240930_182205](https://github.com/user-attachments/assets/493c63a7-5faf-4c37-9694-308697d767d9)

### Contribution

Lots to do in this issue, and totally fine to do more than one PR for the various changes! 🎃 Happy to support as needed and review PRs as they come 😊

indra057:
hello sir can u please assign me this issue so i can get started .. and  if i need any help can i message here?

andrewtavis:
Sure thing, @indra057! Please let us know if you need any assistance and we'll do our best to help 😊

indra057:
sir i didnt find the code exactly where do i need to fix the issue can u please guide me..or give me some more hints...

indra057:
@andrewtavis sir please reply

andrewtavis:
Hi sorry, @indra057. A lot has come up recently between work and another program starting. @angrezichatterbox, would you be able to support here and direct @indra057 to where the change would need to happen for the command bar coloration?

indra057:
sorry @andrewtavis sir i am new to android development and i can't support that much as i didn't even understand the codebase and what is exactly happening i am just looking for simple issues that i can fix 


where  should i change the commands exactly so that i can fix the issue of the coloration of the toolbar

[...]

PR #300, opened 2025-01-18
Implement Data Contract-Based Noun Suggestion with Plural and Gender Handling

angrezichatterbox:
### Contributor checklist

---

### Description

The PR introduces a Data Contract-based implementation for Noun Suggestion, designed to enhance modularity and flexibility in handling language-specific noun forms. This implementation leverages JSON-defined Data Contracts that are processed using a DataContractModel, located in the app's models module. The model enforces the presence of three essential keys within each contract: numbers, genders, and conjugation. Additionally, it assumes six forms of verbs for each conjugation, initialized as empty maps by default if not explicitly defined.
### Plural Suggestion
The plural suggestion logic extracts plural forms from the numbers column in the Data Contract. These plural words are gathered into a list, which is then used to evaluate user input. If the last typed word matches any word in this list, it is identified as plural and marked with the label "PL," regardless of the word's gender. 
### Noun Gender Suggestion
For handling gender information, the implementation employs the GenderDataManager class, which interacts with a SQLite database. This class operates within the constraints of a specific language and utilizes the associated DataContract to organize and interpret data effectively. The GenderDataManager performs checks to ensure the database file's existence and processes gender data based on the structure:
1. If a main gender column is present, it processes data directly from this column.
2. If separate columns for masculine and feminine genders are available, it accommodates this format and processes accordingly.

### Key Assumptions
This implementation relies on several foundational assumptions:
* Every Data Contract must include the numbers, genders, and conjugation keys.
* Six verb forms must be defined under each conjugation. In cases where these forms are missing, they default to empty maps.

### Related issue

-   #135

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -19,6 +19,7 @@ plugins {
     id("de.mannodermaus.android-junit5") version "1.11.2.0"
     id("org.jetbrains.kotlin.plugin.compose") version "2.0.0"
     id("jacoco")
+    kotlin("plugin.serialization") version "1.9.0"
 }
 
 jacoco {
@@ -247,6 +248,7 @@ dependencies {
     api("com.google.code.gson:gson:2.10.1")
     api("com.github.bumptech.glide:glide:4.14.2")
     ksp("com.github.bumptech.glide:ksp:4.14.2")
+    implementation("org.jetbrains.kotlinx:kotlinx-serialization-json:1.6.3")
 }
 
 tasks.register<Copy>("moveFromi18n") {
--- a/app/src/main/assets/data-contracts/de.json
+++ b/app/src/main/assets/data-contracts/de.json
@@ -28,12 +28,12 @@
         },
         "3": {
             "title": "Perfekt",
-            "1": { "ich": "" },
-            "2": { "du": "" },
-            "3": { "er/sie/es": "" },
-            "4": { "wir": "" },
-            "5": { "ihr": "" },
-            "6": { "sie/Sie": "" }
+            "1": { "ich": "[auxiliaryVerb] pastParticiple" },
+            "2": { "du": "[auxiliaryVerb] pastParticiple" },
+            "3": { "er/sie/es": "[auxiliaryVerb] pastParticiple" },
+            "4": { "wir": "[auxiliaryVerb] pastParticiple" },
+            "5": { "ihr": "[auxiliaryVerb] pastParticiple" },
+            "6": { "sie/Sie": "[auxiliaryVerb] pastParticiple" }
         }
     }
 }
--- a/app/src/main/assets/data-contracts/en.json
+++ b/app/src/main/assets/data-contracts/en.json
@@ -1,6 +1,12 @@
 {
     "numbers": { "singular": "plural" },
-    "genders": "NOT_INCLUDED",
+    "genders":  {
+        "canonical": ["NOT_INCLUDED"],
+        "feminines": ["NOT_INCLUDED"],
+        "masculines": ["NOT_INCLUDED"],
+        "commons": ["NOT_INCLUDED"],
+        "neuters": ["NOT_INCLUDED"]
+    },
     "conjugations": {
         "1": {
             "title": "Present",
--- a/app/src/main/assets/data-contracts/en.json
+++ b/app/src/main/assets/data-contracts/en.json
@@ -1,21 +1,42 @@
 {
     "numbers": { "singular": "plural" },
-    "genders": {
-        "canonical": [],
-        "feminines": [],
-        "masculines": [],
-        "commons": [],
-        "neuters": []
-    },
+    "genders": "NOT_INCLUDED",
     "conjugations": {
         "1": {
-            "title": "",
-            "1": "",
-            "2": "",
-            "3": "",
-            "4": "",
-            "5": "",
-            "6": ""
+            "title": "Present",
+            "1": { "I": "simplePresent" },
+            "2": { "you": "simplePresent" },
+            "3": { "he/she/it": "simplePresentThirdPersonSingular" },
+            "4": { "we": "simplePresent" },
+            "5": { "you all": "simplePresent" },
+            "6": { "they": "simplePresent" }
+        },
+        "2": {
+            "title": "Past",
+            "1": { "I": "simplePast" },
+            "2": { "you": "simplePast" },
+            "3": { "he/she/it": "simplePast" },
+            "4": { "we": "simplePast" },
+            "5": { "you all": "simplePast" },
+            "6": { "they": "simplePast" }
+        },
+        "3": {
+            "title": "Perfect",
+            "1": { "I": "presentParticiple simplePast" },
+            "2": { "you": "presentParticiple simplePast" },
+            "3": { "he/she/it": "presentParticiple simplePast" },
+            "4": { "we": "presentParticiple simplePast" },
+            "5": { "you all": "presentParticiple simplePast" },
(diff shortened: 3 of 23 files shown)


angrezichatterbox:
Should I add the license to all the files that I have newly created ?

andrewtavis:
We'll have the new license notice, so let's not worry about it for now and it can be fixed in the PR that changes all the files 😊

andrewtavis:
This is some amazing work here, @angrezichatterbox! So great to see this taking shape 😊 Will be so helpful moving forward 🥳

Wanted to check on the functionality for nouns that have two genders. I'm seeing that `Schild` isn't leading to a gender on German keyboards. That's should be a blue M next to a green N given it's forms. The German word `See` should be a red F next to a blue M as it's lake and ocean and has different genders based on each. I also always check `Leben`, which should be a green N next to an orange PL. We'd talked not showing PL with anything else, but I think I wasn't clear there. For Spanish this isn't needed as we can just say they're plural and don't need to say that they're also feminine or masculine, but then that's also because that language is giving us a non-standard "feminine plural" rather than "plural for a feminine noun" based on the data. For German `Leben` is both neutral and is it's own plural, which is important to indicate to the user as if they try to use the plural command they'll be confused by the response if we don't make it clear that it's both.

How do we want to proceed here? Do you want to expand the noun-gender annotations so that we have the split annotations? This is how it looks on iOS, btw:

Simulator Screenshot - iPhone 16 Pro - 2025-01-18 at 19 59 48

Simulator Screenshot - iPhone 16 Pro - 2025-01-18 at 19 59 10

Let me know, and thanks again for the amazing implementation so far!

angrezichatterbox:
So in German only we have these cases. Like in case of German or any db that has a gender column I could make the Type into a List so that if  its already added to a specific gender it would also have other types of the same word to be present as possible noun type for the word. 

As for the Plural, How should it be handled. Should I make all Plural word shows their gender as well.

andrewtavis:
I'd say that we shouldn't do this. The case for `Leben` is that it's singular and plural, but if we do that for all plurals then people would think that `Hermanas` (sisters in Spanish) would also be a singular word. So if it's in the plural column show plural, if it's in a singular column and a plural show both the gender and plural, and if it's just in a singular column or has a canonical gender and isn't in plural then it's just the gender :)

How does this sound? 😊

angrezichatterbox:
> I'd say that we shouldn't do this. The case for `Leben` is that it's singular and plural, but if we do that for all plurals then people would think that `Hermanas` (sisters in Spanish) would also be a singular word. So if it's in the plural column show plural, if it's in a singular column and a plural show both the gender and plural, and if it's just in a singular column or has a canonical gender and isn't in plural then it's just the gender :)
> 
> How does this sound? 😊

Ok I think I have understood it. I will make the changes and sent in the updated PR by Tuesday or Wednesday.

angrezichatterbox:
I will squash the commit that got duplicated. I had some issues since my git repository got corrupted. Could I get a review on the current design for the dual suggestion @andrewtavis . I am happy to make any cosmetic changes or functional changes to make this PR perfect. It is not yet ready for merge since I have to make some changes to make it work for English keyboard.

andrewtavis:
I brought it down and tried to play around with it on German as that's where a lot of double gender strings are, but noun annotations aren't showing up for the German keyboard at all 🤔

angrezichatterbox:
> I brought it down and tried to play around with it on German as that's where a lot of double gender strings are, but noun annotations aren't showing up for the German keyboard at all 🤔

Does it not come up even for words like `Schild` or `Leben`

angrezichatterbox:
Just wanted to confirm. 

```json
        "1": {
            "title": "",
            "1": "",
            "2": "",
            "3": "",
            "4": "",
            "5": "",
            "6": ""
        }
```

This individual text would be replaces with a structure similar to 

``` json

  "1": {
            "title": "Presente",
            "1": { "yo": "indicativePresentFirstPersonSingular" },
            "2": { "tú": "indicativePresentFirstPersonPlural" },
            "3": { "él/ella/Ud.": "indicativePresentSecondPersonSingular" },
            "4": { "nosotros": "indicativePresentSecondPersonPlural" },
            "5": { "vosotros": "indicativePresentThirdPersonSingular" },
            "6": { "ellos/ellas/Uds.": "indicativePresentThirdPersonPlural" }
        },


```

where we would have Map in the individual keys other than title.

andrewtavis:
> Just wanted to confirm.
> 
> ```json
>         "1": {
>             "title": "",
>             "1": "",
>             "2": "",
>             "3": "",
>             "4": "",
>             "5": "",
>             "6": ""
>         }
> ```
> 
> This individual text would be replaces with a structure similar to
> 
> ```json
>   "1": {
>             "title": "Presente",
>             "1": { "yo": "indicativePresentFirstPersonSingular" },
>             "2": { "tú": "indicativePresentFirstPersonPlural" },
>             "3": { "él/ella/Ud.": "indicativePresentSecondPersonSingular" },
>             "4": { "nosotros": "indicativePresentSecondPersonPlural" },
>             "5": { "vosotros": "indicativePresentThirdPersonSingular" },
>             "6": { "ellos/ellas/Uds.": "indicativePresentThirdPersonPlural" }
>         },
> ```
> 
> where we would have Map in the individual keys other than title.

Yes exactly, @angrezichatterbox :) I can't really think of a better way of making sure that things are readable for the developer to know what goes where and also make sure that we have the label for the conjugation field.

andrewtavis:
As discussed in #161, I invalidated my cache and remade my emulator and all's working better now, @angrezichatterbox 😊 Sorry for the wait here! Glad that it's at least in the docs now :)

Designs are looking really good! ✨ Do you want to add anything else to this PR, or should I do a final check of the code and merge it in? :)

andrewtavis:
Feel free to add in placeholder strings and we can merge :) The contract won't have empty strings, and maybe if the functionality isn't needed for a language we can do `"NOT_INCLUDED"` for the value - like for English noun genders.

andrewtavis:
See cc48f5f for the most recent contracts, @angrezichatterbox :) I think that this can work. We can just check for `"NOT_INCLUDED"` and cancel operations if that's the case.

andrewtavis:
I guess it'd make sense for you to do any changes needed to get the English keyboard working and then we can merge? Feel free to edit `"NOT_INCLUDED"` to a different placeholder or a placeholder object or something like that if it makes using the contract easier! 😊

angrezichatterbox:
I have made a slight change to the contract to be like this.

```json
 "genders":  {
        "canonical": ["NOT_INCLUDED"],
        "feminines": ["NOT_INCLUDED"],
        "masculines": ["NOT_INCLUDED"],
        "commons": ["NOT_INCLUDED"],
        "neuters": ["NOT_INCLUDED"]
    },
```
This was because I assumed that gender would have 5 columns by default where each would be of type List. I hope it is fine.

angrezichatterbox:
Would it make sense to make a workflow that would run whenever the data contract are changed so that we would get to know whether we have to make changes to the contract loader ?

andrewtavis:
Ideally the loader would never change, but then the chances of that are quite low, so yes some kind of testing would be needed. That's a different issue though for sure.

andrewtavis:
With the above changes and us making a new issue for the loader check, we're good for a final check here and merging it in?

angrezichatterbox:
> With the above changes and us making a new issue for the loader check, we're good for a final check here and merging it in?

Yes, It's good for a final check.

andrewtavis (approved):
All's looking great here, @angrezichatterbox! Thanks for the first implementation of the data contracts! 🚀 I'll make the issue that we discussed :)

PR #305, opened 2025-01-31
Fix Command Bar Spacebar Input & Implement Plural Command

smalik2811:
### Contributor checklist

---

### Description
This Pull Request addresses two key issues:

1. **Fixes incorrect spacebar behavior in command mode:**
    - Previously, pressing the spacebar in command mode would incorrectly add `spaces` to the main input connection instead of the command bar input.
    - This fix ensures that `spaces` are correctly added to the command bar, enabling users to enter multi-word commands as expected. This is crucial for supporting commands with multi-word lexemes.

2. I**mplements the plural command functionality:**

    This feature allows users to insert a plural form of the noun entered in the command bar using the Enter key.

    - The plural form of the noun is first retrieved from the language database using Data Contracts.
    - Then, the plural form of the noun is inserted into the input connection.
    - Finally, the keyboard is switched to Idle mode.

### Related issue
-   #272

### DIFF
--- a/app/src/main/java/be/scri/helpers/DatabaseHelper.kt
+++ b/app/src/main/java/be/scri/helpers/DatabaseHelper.kt
@@ -6,6 +6,7 @@
 
 package be.scri.helpers
 
+import DataContract
 import android.content.Context
 import android.database.sqlite.SQLiteDatabase
 import android.database.sqlite.SQLiteOpenHelper
@@ -60,5 +61,15 @@ class DatabaseHelper(
         dbManagers.pluralManager.checkIfWordIsPlural(
             language,
             getRequiredData(language),
+        )
+
+    fun getPluralRepresentation(
+        language: String,
+        noun: String,
+    ): Map<String, String?> =
+        dbManagers.pluralManager.queryPluralRepresentation(
+            language,
+            getRequiredData(language),
+            noun,
         )
 }
--- a/app/src/main/java/be/scri/helpers/KeyboardDBHelper/PluralFormsManager.kt
+++ b/app/src/main/java/be/scri/helpers/KeyboardDBHelper/PluralFormsManager.kt
@@ -18,10 +18,27 @@ class PluralFormsManager(
         }
 
         val dbFile = context.getDatabasePath("${language}LanguageData.sqlite")
-        val pluralForms = jsonData.numbers.values.toList()
+        val pluralForms = jsonData!!.numbers.values.toList()
         Log.d("MY-TAG", "Plural Forms: $pluralForms")
 
         return queryPluralForms(dbFile.path, pluralForms)
+    }
+
+    fun queryPluralRepresentation(
+        language: String,
+        jsonData: DataContract?,
+        noun: String,
+    ): Map<String, String?> {
+        if (jsonData?.numbers?.values.isNullOrEmpty()) {
+            Log.e("MY-TAG", "JSON data for 'numbers' is null or empty.")
+            return mapOf()
+        }
+
+        val dbFile = context.getDatabasePath("${language}LanguageData.sqlite")
+        val pluralForms = jsonData!!.numbers.values.toList()
+        val singularForms = jsonData.numbers.keys.toList()
+
+        return queryPluralForms(dbFile.path, pluralForms, singularForms, noun)
     }
 
     private fun queryPluralForms(
@@ -36,8 +53,30 @@ class PluralFormsManager(
                 processPluralFormsCursor(cursor, pluralForms, result)
             }
         }
-
+        db.close()
         return result
+    }
+
+    private fun queryPluralForms(
+        dbPath: String,
+        pluralForms: List<String>,
+        singularForms: List<String>,
+        noun: String,
+    ): Map<String, String?> {
+        val result = mutableListOf<String>()
+        val db = SQLiteDatabase.openDatabase(dbPath, null, SQLiteDatabase.OPEN_READONLY)
+        val queryBuilder = StringBuilder("SELECT * FROM nouns WHERE")
+        val placeholders = singularForms.joinToString("OR ") { "$it = ?" }
+        queryBuilder.append(" $placeholders;")
+        val selectionArgs = Array(singularForms.size) { noun }
+
+        db.use { database ->
+            database.rawQuery(queryBuilder.toString(), selectionArgs)?.use { cursor ->
+                processPluralFormsCursor(cursor, pluralForms, result)
+            }
+        }
+        db.close()
+        return pluralForms.zip(result).toMap()
     }
 
     private fun processPluralFormsCursor(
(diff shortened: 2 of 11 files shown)


andrewtavis:
Functionality wise it's working really great, @smalik2811 😊 One thing that I'm realizing I forgot to mention: Could we also add a space after the returned word? That way the user can directly start typing the rest of what they're trying to say :)

@angrezichatterbox, would you be able to check to code implementation? Also something to note, I wasn't able to pick up the changes here without creating a new device. I invalidated all caches and it still wasn't working. Would be great if we could figure out why this is happening so that we can expand out the contributing guide a bit more with better suggestions of what to do :)

smalik2811:
@andrewtavis Sure, adding a space after the word is a nice idea. I will make the required changes. 😊

angrezichatterbox:
> @angrezichatterbox, would you be able to check to code implementation? Also something to note, I wasn't able to pick up the changes here without creating a new device. I invalidated all caches and it still wasn't working. Would be great if we could figure out why this is happening so that we can expand out the contributing guide a bit more with better suggestions of what to do :)

I will look into the code implementation by evening and also look into the changes causes error for the old device.

andrewtavis:
Thanks @smalik2811 and @angrezichatterbox! Looking forward to bringing this in :)

angrezichatterbox:
The code looks good to me. One thing I was considering is whether it would be better to prepare a HashMap at the beginning when the keyboard opens and then query from that. However, I see that this feature is only used in a specific mode, so I assume that's why you chose to query it on user interaction rather than preloading it at the start. Was this decision made to avoid slowing down the keyboard's opening time, especially as more features get added over time?

@andrewtavis In Scribe-iOS, is querying handled for all modes at the beginning by preparing a HashMap, or is it done dynamically while the user types? Or is it a combination of both, depending on the situation?

So far, we’ve been preparing a HashMap for data  and then retrieving values dynamically based on user input for gender suggestions, emoji suggestions, and is plural auto-suggestions as I assumed this would be much faster.

angrezichatterbox:
> Also something to note, I wasn't able to pick up the changes here without creating a new device. I invalidated all caches and it still wasn't working.

I was able to pick up the changes this time around. However I have faced this issue recently especially with these keyboard issues. I would look into  and check for the possible reason and fix to it.

smalik2811:
> The code looks good to me. One thing I was considering is whether it would be better to prepare a HashMap at the beginning when the keyboard opens and then query from that. However, I see that this feature is only used in a specific mode, so I assume that's why you chose to query it on user interaction rather than preloading it at the start. Was this decision made to avoid slowing down the keyboard's opening time, especially as more features get added over time?

Presently in the database, the median number of records for the `nouns` table is `30K` and the maximum goes to `150K`. Therefore, preloading the entire dataset could lead to high memory usage and could affect the user experience. Since the `Scribe-keyboard` service only runs along with some other activity, we should keep the memory usage on the lower side.

While the features like gender suggestions, emoji suggestions, and is-plural auto-suggestions are fired for every single word typed, therefore it's a good idea to store them in the hashmap because querying database for every single word could be very slow. However the features like `plural`, `translate`, `conjugate` would be less often used therefore querying them when needed seems a better approach to me.

Let me know if you'd prefer the HashMap approach, and I would make the necessary changes 😊

angrezichatterbox:
> While the features like gender suggestions, emoji suggestions, and is-plural auto-suggestions are fired for every single word typed, therefore it's a good idea to store them in the hashmap because querying database for every single word could be very slow. However the features like `plural`, `translate`, `conjugate` would be less often used therefore querying them when needed seems a better approach to me.

So currently those features are only required when the user presses the space button. The only reason I assumed we would need it is because the user would be often press the space while typing. I thought for each of those query it would be quite resource intensive. But when you think about it this approach also sounds nice for that specific use case like noun type suggestion and in long term feels better. Love to hear your thoughts on it @andrewtavis.

andrewtavis:
One question I'd have is how the work here would be done? Would we finish the issue as is and then convert over to a hashmap later? That would be more appropriate to me so we can get the functionality in and then optimize it later in another issue :) I'm not sure on full vs. hybrid right now, but we can discuss in the new issue? If this makes sense, then would one of you want to summarize the current discussion in a new issue?

angrezichatterbox:
> then would one of you want to summarize the current discussion in a new issue?

Ok it would make sense to summarize these into a new issue. Currently here We use a Query whenever the user enters a text and clicks on the enter icon in the plural mode to get the plural word of the one entered. I guess it makes sense to use it here as it is a specific mode. The discussion was that would Hashmap be better in this situation or not. Also whether we should use the approach that he has used in the PR for the other suggestions.

smalik2811:
@andrewtavis, @angrezichatterbox  In my latest commit, which was made to to add an extra space character after the plural word as suggested by Andrew, I missed some edge cases. I will shortly add another commit to handle those edge cases.

andrewtavis:
Thank you both for the discussion and work here! Sorry for needing to pass things to a new issue. Am at a conference yesterday and today, so it's easier for me to compartmentalize the various discussions here :)

smalik2811:
@angrezichatterbox, For the requirement of adding an extra `space` character after inserting the text from plural command, I am unable to get the noun type suggestion working for the inserted word. I have tried to explain the issue with a demo in the video attached below.


https://github.com/user-attachments/assets/7a7faa15-b7b9-455f-b8d6-1902a4b38486



Is it okay if I just push these changes, and the suggestion bug could be fixed in some other issue?

andrewtavis:
I think it's fine to send the current state and we can open a new issue, @smalik2811 :) In iOS we basically are triggering all of this on space, but it's obviously not a very efficient way of doing this. Let's get this in and then we can figure out how best to call these functions 😊

smalik2811:
Great, I have pushed the latest commit, it contains a fix that prevents addition of `space` character if the plural command returns an empty string in case the noun is not available in the database. 😊

angrezichatterbox:
After this is merged we have to look into this issue as well as the issue where the output from the modes should also make command mode go to false and the enter key should go back to the previous icon rather than the play icon.

smalik2811:
@angrezichatterbox Additionally a related issue: When the noun type suggestion is visible like `PL` in the idle mode and the command-select mode is activated, the font size of the translate button does not update.

angrezichatterbox:
> @angrezichatterbox Additionally a related issue: When the noun type suggestion is visible like `PL` in the idle mode and the command-select mode is activated, the font size of the translate button does not update.

Thanks :) . We would have to look into the mode switching once again and the various changes each mode switch should cause. Since some of the modes changes don't reflect on the keyboard when a certain annotation is shown it still remains.

andrewtavis (approved):
All's looking great on my end, @smalik2811! Thanks for the quick reaction to the added space request as well as the work to fix the spacebar functionality 😊🚀

andrewtavis:
From here it would be great if we could open issues for the following:
- Showing the noun annotation after the Plural command
- The discussion for whether we want to use hashmaps for commands and to what degree with the work for that then being done based on the decision in the issue :)

andrewtavis:
Happy to make either or both of the above issues. Just let me know!

Issue #149, opened 2024-09-23
Add the menu option to set Translation source


angrezichatterbox:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Android/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-Android's [Code of Conduct](https://github.com/scribe-org/Scribe-Android/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

Currently the individual language settings doesn't allow the user to set the translation source for each language keyboard settings. The issue aims to add the ability to set translation source language for each language keyboard separately. This could be stored in the shared preferences for now for each language separately similar to how it is being done for period_on_double_tap. 

 A visual reference of how this option would look is given below. 
 
 <img width="385" alt="Screenshot 2024-09-24 at 12 39 00 AM" src="https://github.com/user-attachments/assets/46d55d8d-db09-4472-af74-29133ffc7a49">

### Contribution

I would love to work on this and am more than happy to help anyone fix this issue :)

ThousandYearGrind:
@henrikth39, Hi, does this work?

andrewtavis:
Hey @ThousandYearGrind 👋 Quick note that this issue is now just to add the menu option that selects from a list, so at the end of it we want the menu option that's shown in the issue text and the interface like [this in the Figma designs](https://www.figma.com/design/c8945w2iyoPYVhsqW7vRn6/scribe_public_designs?node-id=544-1527&t=xkn0oeQkAyfkQzxU-1). Actually switching the translation language will be handled by #265 once this issue is finished - i.e. this issue is for the interface and #265 is for the functionality.

ThousandYearGrind:
Ahhh, thank you. I was scratching my head over that for a while.

ThousandYearGrind:
@andrewtavis How should I approach translations? Can I just run "Translation source language" through google translate or should this be done later?

andrewtavis:
We'll have local translations for this, @ThousandYearGrind :) This issue is just for the menu option to select the base language.

ThousandYearGrind:
Hi @andrewtavis, where would be a good place to ask questions about code-specific questions? I've been trying to figure out how to do something for a while.

andrewtavis:
@ThousandYearGrind, asking them here would be good, that way it's easy for us to link to files :)

ThousandYearGrind:
[...]

PR #330, opened 2025-03-03
Added translation source language option to language setting screen

ThousandYearGrind:
### Contributor checklist

---

### Description

Added translation source language option to language settings screen
The actual selection does not work (just a clickable ui item)

I ran my code in Android Studio and saw the option in Settings > Select Installed Keyboard > English
I also ran the ./gradlew lintKotlin detekt test to check my code

### Related issue

-   #149

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -194,7 +194,7 @@ dependencies {
     implementation("androidx.recyclerview:recyclerview:1.3.2")
     implementation("androidx.cardview:cardview:1.0.0")
     implementation("androidx.viewpager2:viewpager2:1.1.0")
-    implementation("com.google.android.play:core:1.10.0")
+    implementation("com.google.android.play:core:1.10.3")
     implementation("androidx.navigation:navigation-compose:2.6.0")
 
     // Jetpack Compose BOM
--- a/app/src/main/java/be/scri/App.kt
+++ b/app/src/main/java/be/scri/App.kt
@@ -31,6 +31,7 @@ import be.scri.ui.common.bottombar.ScribeBottomBar
 import be.scri.ui.screens.InstallationScreen
 import be.scri.ui.screens.LanguageSettingsScreen
 import be.scri.ui.screens.PrivacyPolicyScreen
+import be.scri.ui.screens.SelectTranslationSourceLanguageScreen
 import be.scri.ui.screens.ThirdPartyScreen
 import be.scri.ui.screens.WikimediaScreen
 import be.scri.ui.screens.about.AboutScreen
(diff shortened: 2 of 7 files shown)


angrezichatterbox:
Hey @ThousandYearGrind.

The issue actually wants you to show the different languages on clicking the select language menu option and then select the specific language and store it in shared preferences that could be accessed later on for different languages differently. 

Would you like to do that in the same PR

ThousandYearGrind:
Hi @angrezichatterbox 
I tried to do that initially but did not know how to open up a separate menu for language selection. Maybe I can come back to it after I learn more, but for now it would probably be better to open it up so other people who do know how to do it could work on it.

angrezichatterbox:
Should I make changes within the same PR for adding the individual languages or a menu or should a separate issue be made for it @andrewtavis

andrewtavis:
I think we'd be good to make changes within this PR and resolve the issue, @angrezichatterbox :) Thanks for checking!

angrezichatterbox:
Should The translation selection be a separate page or should an overlay with the  options be present.

andrewtavis:
I think the plan was a separate page so it's a bit consistent with going to sub menu pages, but then I haven't actually thought of other options 🤔 I think a popover would be better than a dropdown. See the [designs here](https://www.figma.com/design/c8945w2iyoPYVhsqW7vRn6/scribe_public_designs?node-id=513-1616&t=uLCHd4YfEB2OLbPA-1). If we want to switch to a popover, then we should think of a different interaction icon for these options than `>`, as that indicates that we're going to a sub-page.

Any thoughts on this, @angrezichatterbox? :)

angrezichatterbox:
> I think the plan was a separate page so it's a bit consistent with going to sub menu pages, but then I haven't actually thought of other options 🤔 I think a popover would be better than a dropdown. See the [designs here](https://www.figma.com/design/c8945w2iyoPYVhsqW7vRn6/scribe_public_designs?node-id=513-1616&t=uLCHd4YfEB2OLbPA-1). If we want to switch to a popover, then we should think of a different interaction icon for these options than `>`, as that indicates that we're going to a sub-page.
> 
> Any thoughts on this, @angrezichatterbox? :)

I think subpages would be good if we would have a lot of language at any point and the consistency would be maintained across the app.

andrewtavis:
Good point that scrolling a popover would be really annoying 😊 Let's stick with sub pages as is in the designs then!

angrezichatterbox:
It is ready for review :)

andrewtavis:
Nice @angrezichatterbox! I'll get to this later today :)

andrewtavis (approved):
All's functioning really well, @angrezichatterbox! And thanks for the great start here, @ThousandYearGrind 😊 Glad to have the menu for keyboards so well set up 🥳

Issue #316, opened 2025-02-26
Standardize the color of the Scribe key, command keys and return key in command mode


andrewtavis:
### Terms

- [x] I have searched [open and closed design issues](https://github.com/scribe-org/Scribe-Android/issues?q=is%3Aissue+label%3Adesign)
- [x] I agree to follow Scribe-Android's [Code of Conduct](https://github.com/scribe-org/Scribe-Android/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

As of now the Scribe key and the command keys are different shades of blue. Each of them should be Scribe blue for the given color mode. Furthermore the return key in command mode should have a matching color with the other keys mentioned. We can see all of these keys via:

- The Scribe key is at the top left of the keyboard interface
- Clicking the Scribe key shows the command keys (there as of now a different shade of blue)
- Clicking one of the command keys switches the keyboard to command mode and the return key turns blue and has a play button icon
    - As of now it's the same color as the command keys, and should be matched

As of writing I'm not 100% sure which color is correct. I think it's the Scribe key, but we can figure out the exact colors once someone is assigned :)

### Contribution

Happy to support someone on this and review once a PR is opened! ✈

janiceferns02:
@andrewtavis Can you assign this to me? I see that part of it is already resolved

andrewtavis:
Thanks for picking this up, @janiceferns02! Think it makes total sense for you to work on this 😊

andrewtavis:
Closed by #143 🚀 Thanks so much for your efforts on Scribe-Android, @janiceferns02!

PR #340, opened 2025-03-20
Set Return key color in command mode to match the scribe key in dark m…

janiceferns02:
…ode.

### Contributor checklist

---

### Description

Set Return key color in command mode to match the scribe key and the command keys in the dark mode. Tested this by switching dark mode in scribe app and the device dark/light modes.

### Related issue
-   #316


andrewtavis (approved):
All's looking great now, @janiceferns02! Thanks so much 😊 I'll make another issue right now for the size of the Scribe key icon and the `X` when it's used to cancel the command mode. Maybe you'd want to work on that, and then we can move to something else? Really is great to have your dedication to the project!

Issue #337, opened 2025-03-19
Add coverage report to PRs


angrezichatterbox:
### Terms

- [x] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Android/issues?q=is%3Aissue+label%3Afeature)
- [x] I agree to follow Scribe-Android's [Code of Conduct](https://github.com/scribe-org/Scribe-Android/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

Currently, Scribe-Android does not generate or display code coverage reports in PRs. Adding a coverage report will help track test coverage, improve code quality, and encourage better testing practices. This issue aims to integrate a code coverage tool and display reports in PRs. It would also help us understand what all tests are reminding to be written. 

### Contribution

I would be working on this :)

andrewtavis:
Closed by #350 ✅🚀 Amazing work here, @angrezichatterbox! Another case where Android's ahead of iOS 😊

PR #350, opened 2025-03-28
Add Coverage Report for  PR

angrezichatterbox:
### Contributor checklist

---

### Description

The PR introduces coverage report for the PR.

### Related issue

-   #337

### DIFF
--- a/.github/workflows/android_ci.yml
+++ b/.github/workflows/android_ci.yml
@@ -53,15 +53,9 @@ jobs:
             - name: Get coverage
               run: ./gradlew koverReport
           
-            - name: Parse Coverage Report
-              run: |
-                sudo apt-get install -y libxml2-utils  # Install xmllint
-                echo "Coverage Summary"
-                echo "---------------------------"
-                echo "Line Coverage: $(xmllint --xpath 'string(//counter[@type="LINE"]/@covered)' app/build/reports/kover/xml/report.xml)/$(xmllint --xpath 'string(//counter[@type="LINE"]/@missed)' app/build/reports/kover/xml/report.xml)"
-                echo "Branch Coverage: $(xmllint --xpath 'string(//counter[@type="BRANCH"]/@covered)' app/build/reports/kover/xml/report.xml)/$(xmllint --xpath 'string(//counter[@type="BRANCH"]/@missed)' app/build/reports/kover/xml/report.xml)"
-                echo "Instruction Coverage: $(xmllint --xpath 'string(//counter[@type="INSTRUCTION"]/@covered)' app/build/reports/kover/xml/report.xml)/$(xmllint --xpath 'string(//counter[@type="INSTRUCTION"]/@missed)' app/build/reports/kover/xml/report.xml)"
-                echo "---------------------------"
+            - name: Run File-wise Coverage Parser Script
+              run: python3 scripts/coverage_parser.py
+        
             
       
 
--- a/.github/workflows/android_ci.yml
+++ b/.github/workflows/android_ci.yml
@@ -36,28 +36,50 @@ jobs:
               run: ./gradlew test
               
     coverage:
-        runs-on: ubuntu-latest
-        name: Run coverage
-        steps:
-            - name: Checkout repository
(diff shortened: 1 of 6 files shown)


angrezichatterbox:
I would like to know should all the individual file have coverage percentage shown next to them or should I make it directory based. Like helpers directory and so on.

From Jacoco I would get package wise coverage report in the main document which would include the coverage for each package.

angrezichatterbox:
Alicia has setup the basis for this long back. I had to add some sample tests for going through the classes to make it work properly. I also noticed that the repository doesn't allow git actions to add comments. So should I make the output from coverage tests run more clean.

angrezichatterbox:
Some minor changes are required. I will make it and comment in here when it is ready for review :)

andrewtavis:
I guess that a massive coverage report for all the files being commented in would be a bit much, so keeping it in the action output and checking a threshold to fail against if it's below it would be enough? What do you think, @angrezichatterbox? :)

angrezichatterbox:
> I guess that a massive coverage report for all the files being commented in would be a bit much, so keeping it in the action output and checking a threshold to fail against if it's below it would be enough? What do you think, @angrezichatterbox? :)

So I would want to make it fail if the overall  coverage is less than a certain value right ?

andrewtavis:
Exactly, and for now just use the current coverage rounded to the highest 5%. We're likely not that high anyway, so hopefully we can at least do 5%, but if not let's just set it to 0 and increase it with new testing issues as they're finished ✅

angrezichatterbox:
> Exactly, and for now just use the current coverage rounded to the highest 5%. We're likely not that high anyway, so hopefully we can at least do 5%, but if not let's just set it to 0 and increase it with new testing issues as they're finished ✅

I think it is currently 0 or 0.15 %. 

Jacoco bot currently return the code coverage of the entire codebase and the changed file. The thing is git actions bot is not able to comment it right now which I am not sure why. I would made the output more clear and will bring in the updated changes by tomorrow.

andrewtavis:
Ok :) It's also totally fine if we need to check the workflow for the coverage percent, but I trust you to find the appropriate solution 😊

angrezichatterbox:
So this is what I have done. 

I am returning the coverage percentage current. I have not made it fail since currently it's 0 percent. I have also returned the coverage report as an artifact.

andrewtavis:
Is it possible to also get the coverage report to be printed in the terminal of the job so we can check it directly, @angrezichatterbox? I checked the courage report workflow and did see the artifact being uploaded, but it would also be nice to see the results directly :)

angrezichatterbox:
> Is it possible to also get the coverage report to be printed in the terminal of the job so we can check it directly, @angrezichatterbox? I checked the courage report workflow and did see the artifact being uploaded, but it would also be nice to see the results directly :)

Okay I will add that change and sent in the updated PR.

angrezichatterbox:
Currently I am getting an XML, I could also have a html doc. The idea which comes to my mind is to parse this XML doc and print it as a proper output using a python script. This is what comes to my mind. The existing git actions present in marketplace are now outdated which does the same purpose. Others just give for changed file and the overall coverage.

andrewtavis:
Doing a Python or Kotlin file to parse the results would be totally fine, @angrezichatterbox! Maybe doing a Python file instead of a Swift file for https://github.com/scribe-org/Scribe-iOS/pull/532 might also make sense as for some reason the Swift file isn't working 🤔 Feel free to add that in and then we'll be good to go!

andrewtavis:
Also feel free to just use a Python file. I honestly think that that might be the problem with my workflow for iSO 🤔

angrezichatterbox:
I have parsed the report and print it on the actions console. It is ready for review :)

andrewtavis (approved):
Amazing work here, @angrezichatterbox! Great that we're getting this done so early in the app's history, and also great that we have it done right before a sync and can discuss next steps 😊 Really awesome that we'll be able to be so directed in our test development, and from there we can work towards test-driven development 🥳

PR #399, opened 2025-05-17
Fix Translate button text color in SELECT_COMMAND state

### LINKED ISSUE #397: Typing text in `SELECT_COMMAND` causes Translate Button Text to turn black
### Terms

- [x] I have searched all [open bug reports](https://github.com/scribe-org/Scribe-Android/issues?q=is%3Aopen+is%3Aissue+label%3Abug)

### Behavior

When the Scribe key is clicked and the user enters the `SELECT_COMMAND` state, the text color of the Translate button changes to black as the user begins typing. This behavior is inconsistent with the expected UI, where the Translate button's text color should remain unchanged regardless of user input. 

The issue is to make sure the Translate Button text color remain the same when the user types in `SELECT_COMMAND` state.

### Device type

Pixel 9 Pro API 35

### Versions

_No response_

bhanu-dev82:
### Contributor checklist

---

### Description

**Problem:**
When the Scribe key is clicked and the user enters the `SELECT_COMMAND` state, the text color of the "Translate" button incorrectly changes to black, especially noticeable as the user begins typing. This behavior is inconsistent with the expected UI, where the "Translate" button's text color should remain unchanged (i.e., respect the current dark/light theme) regardless of user input in this state.

https://github.com/user-attachments/assets/b550c8f7-472f-44fd-aa39-6037482d78bb

**Solution:**
The root cause was identified as an explicit call to `binding.translateBtn.setTextColor(Color.BLACK)` within the `setupSelectCommandView()` method. This call occurred *after* `setupCommandBarTheme(binding)` had already set the appropriate theme-based text color, effectively overriding it.

This PR removes the erroneous `setTextColor(Color.BLACK)` line. As a result, the "Translate" button's text color will now correctly adhere to the theme (e.g., white in dark mode, dark grey in light mode) as determined by the `setupCommandBarTheme()` function when in the `SELECT_COMMAND` state.

https://github.com/user-attachments/assets/f6e35039-b910-4646-ad0c-7a4807e4a92a

-   Fix #397

### DIFF
--- a/app/src/main/java/be/scri/services/GeneralKeyboardIME.kt
+++ b/app/src/main/java/be/scri/services/GeneralKeyboardIME.kt
@@ -699,7 +699,6 @@ abstract class GeneralKeyboardIME(
         binding.separator5.visibility = View.GONE
         binding.separator6.visibility = View.GONE
         setupCommandBarTheme(binding)
-        binding.translateBtn.setTextColor(Color.BLACK)
         binding.scribeKey.setOnClickListener {
             currentState = ScribeState.IDLE
             Log.i("MY-TAG", "IDLE STATE")


andrewtavis (approved):
Confirmed on by running the emulator that we don't have the text color changing anymore :) Thanks for the quick fix, @bhanu-dev82! 🚀

Issue #174, opened 2024-10-07
app crashing at startup on api 30


MostafaMohamed2002:
### Terms

- [X] I have searched all [open bug reports](https://github.com/scribe-org/Scribe-Android/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow Scribe-Android's [Code of Conduct](https://github.com/scribe-org/Scribe-Android/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

when trying to launch the app on api 30 it crashes at startup 
```17:10:07.321  E  FATAL EXCEPTION: main
                 Process: be.scri.debug, PID: 11099
                 java.lang.RuntimeException: Unable to start activity ComponentInfo{be.scri.debug/be.scri.activities.SplashActivity}: java.lang.UnsupportedOperationException: Can't convert value at index 1 to color: type=0x1
                 	at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:3449)
                 	at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:3601)
                 	at android.app.servertransaction.LaunchActivityItem.execute(LaunchActivityItem.java:85)
                 	at android.app.servertransaction.TransactionExecutor.executeCallbacks(TransactionExecutor.java:135)
                 	at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:95)
                 	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2066)
                 	at android.os.Handler.dispatchMessage(Handler.java:106)
                 	at android.os.Looper.loop(Looper.java:223)
                 	at android.app.ActivityThread.main(ActivityThread.java:7656)
                 	at java.lang.reflect.Method.invoke(Native Method)
                 	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:592)
                 	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:947)
                 Caused by: java.lang.UnsupportedOperationException: Can't convert value at index 1 to color: type=0x1
                 	at android.content.res.TypedArray.getColor(TypedArray.java:533)
                 	at android.app.Activity.onApplyThemeResource(Activity.java:5104)
                 	at android.view.ContextThemeWrapper.initializeTheme(ContextThemeWrapper.java:216)
                 	at android.view.ContextThemeWrapper.setTheme(ContextThemeWrapper.java:147)
[...]

PR #402, opened 2025-05-21
fix startup crash on API < 31

linreal:
### Contributor checklist

---

### Description

Colors in app theme were referencing other colors, which were available only for API level > 30, leading to crash on startup
There is no actual need to have values-v31 colors in app, so I've moved them to base colors.xml, replacing incorrect references

### Related issue

-   #174


andrewtavis (approved):
Confirmed that all's working on API 30 on this branch while it's not on main 😊 Thanks for two amazing first contributions, @linreal! Really quite the start 🥳

Please let us know if there's another issue that would be of interest!

Issue #447, opened 2025-07-26
Auto-capitalization after a full stop


Nishthajain7:
### Terms

- [x] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Android/issues?q=is%3Aissue+label%3Afeature)
- [x] I agree to follow Scribe-Android's [Code of Conduct](https://github.com/scribe-org/Scribe-Android/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

Entering a full stop (.) followed by a space ( ) should trigger auto-capitalization of the next word. This is the standard behavior in most mobile keyboards to assist with sentence formatting.

PR #474, opened 2025-08-13
Auto capitalisation after punctuation. 

Wreck-X:
### Contributor checklist

---

### Description

This update checks for punctuation when normal space is pressed to ensure that the next key is capitalised

### Related issue

-   #447


andrewtavis:
Thanks so much for the PR, @Wreck-X! 😊 We'll try to get to the review in the coming days :)

angrezichatterbox:
I assume this is supposed to work like If we type a few letters and then the space the keyboard would be capitalized right. 
I am currently not able to do that.

angrezichatterbox:
This is what I did 

1. I typed some text. 
2. Put a full stop 
3. Then entered a space. 

After this the keyboard is supposed to be capitalized right.

Wreck-X:
It is yes, lemme check myself I might've missed a few commits

PR #481, opened 2025-08-31
fix: standardize command bar

### LINKED ISSUE #315: Standardization of command bar command directions
### Terms

- [x] I have searched [open and closed design issues](https://github.com/scribe-org/Scribe-Android/issues?q=is%3Aissue+label%3Adesign)

### Description

There's a bit of work that needs to be done to finalize the command bar directions and placeholders. You can see the directions and placeholders by loading up a Scribe keyboard and then clicking the Scribe key at the top left of the keyboard. From there select a command. Specifically:
- The translate placeholder needs to have `:` after the `SOURCE_ISO -> TARGET_ISO` part of the placeholder
- The conjugate placeholder similarly needs `:` after the localized word "Conjugate" in the given language
- The command bar cursor character `│` should be black even if the user hasn't typed anything
- The horizontal distance between the command bar cursor character and the rest of the text should be reduced

### Contribution

Happy to support with the work here and review when a PR is opened! 😊

DeleMike:
### Contributor checklist

---

### Description

Basically, I removed some code that was causing the entered text to start close to the divider.

### Related issue

-   closes #315


DeleMike:
Thanks @andrewtavis!

I will work on the issue with `translate` part. 

For the cursor, first I removed the main cursor and left the divider. But if we want that approach to work, then we need to use the `EditText`'s cursor. And by default, the Android System makes this blinking, and I don't know yet if it is customisable. If it is, it will definitely be a hack and might not be sustainable. 

I will check for more info. 
I will bring Gautham here. @angrezichatterbox, what do you think about this, please?

DeleMike:
The latest commit corrects the translation prompt. 

---

I am still researching what can be done for the previous comment I made. But I think having the cursor is the best way to have the experience you described, and we cannot make the vertical divider(|) do the job. Questions like allowing the user to tap in the middle of a word to edit a letter would arise. For example, the user intends to write "bookman" but writes "boskman", instead of clearing out the whole of "kman", they can just take the cursor to the right of 's' and edit to the right character.

This use case states why the cursor is the best here and why having no cursor is not too good. 

What do you think?

andrewtavis:
Hey @DeleMike 👋 Checking here, is there any further work required here, or are we good to merge? Goal here now based on our discussion in the sync is that we want to remove the blue cursor and make sure that all the commands have a colon `:` at the end of them :) We can then finalize cursor behavior in #486 😊

andrewtavis:
Starting the review for this here now, @DeleMike :) Thanks for merging in main!

Issue #483, opened 2025-09-04
No suggestions are displayed for words without linguistic suggestions


Nishthajain7:
### Terms

- [x] I have searched all [open bug reports](https://github.com/scribe-org/Scribe-Android/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [x] I agree to follow Scribe-Android's [Code of Conduct](https://github.com/scribe-org/Scribe-Android/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

Currently, when a word does not have any linguistic suggestions, no suggestions are displayed at all. The updateAutoSuggestText function relies on the presence of linguistic suggestions or word suggestions to show results, but the fallback path is not triggered correctly for words without linguistic data.

### Device type

All devices

PR #484, opened 2025-09-05
Fix auto suggest

Nishthajain7:
### Contributor checklist

---

### Description

- Verified that words with valid linguistic data show both plural/gender and the suggestions correctly.
- Words without linguistic data now correctly display word-based suggestions instead of showing nothing.
- Added a unit test for processLinguisticSuggestions()

### Related issue

-   #483


andrewtavis:
Thanks for the extra commit here, @Nishthajain7! This will definitely help the user experience :)

PR #489, opened 2025-09-19
fix(suggestions)  make initial suggestions clickable

### LINKED ISSUE #487: Initial suggestion does not get added to input on click
### Terms

- [x] I have searched all [open bug reports](https://github.com/scribe-org/Scribe-Android/issues?q=is%3Aopen+is%3Aissue+label%3Abug)

### Behavior

Currently when there are no auto suggestions available for the word that the user has typed. A set of auto suggestions would appear on the suggestions as seen below. 

Image

When we click on these we are not able to add them to the input connection. This issue is only with the hardcoded auto suggestions and the auto suggestions generated from the word are added to the input. 

### Device type

Pixel 9 Pro API 35 

### Versions

_No response_

bhanu-dev82:
### Contributor checklist

---

### Description

The initial/hardcoded suggestions were not being added to the input
field on click. This was because the `disableAutoSuggest()` function,
which is called when the keyboard is in an idle state, was explicitly
setting the `onClickListener` for the suggestion buttons to null.

This commit attaches a new `OnClickListener` to each default
suggestion button. The listener's logic commits the button's text
to the current input connection, mirroring the behavior of the dynamic,
next-word suggestions.

Closes #487


andrewtavis (approved):
All's working really well here, @bhanu-dev82! Another great addition to the usability of the app 😊 Getting so close! 🚀

Issue #493, opened 2025-09-26
Improve keyboard name in selection


andrewtavis:
### Terms

- [x] I have searched [open and closed design issues](https://github.com/scribe-org/Scribe-Android/issues?q=is%3Aissue+label%3Adesign)
- [x] I agree to follow Scribe-Android's [Code of Conduct](https://github.com/scribe-org/Scribe-Android/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

After work in #472 where we were trying to order the languages as they appear in the UI we discovered a small bug in the UI. When the user is selecting a keyboard they now are seeing the language name in lowercase:

<img width="369" height="300" alt="Image" src="https://github.com/user-attachments/assets/6be4d6eb-9cd2-40db-bdc9-3b2c675ba01d" />

Ideally we'd find a way to uppercase the language, or if this is not possible we need to find a way to switch it to a string that we can uppercase from the Scribe-i18n project.

### Contribution

Happy to support and review a PR for this! 😊

Akshaykomar890:
Can I work on this ?

Akshaykomar890:
Hi @andrewtavis 

I've been testing this issue and found something interesting:

Original code:
android:imeSubtypeLocale="Deutsch"` → showed "deutsch" (lowercase)

What I tried to fix it:
android:imeSubtypeLocale="de"` → showed "Germany" (translated to English)

What this means:
Android reads our custom "Deutsch" setting but forces it to lowercase display
When I tried using standard locale codes like "de", it translates to the system language instead.

So no matter what we put, Android either forces lowercase or translates it. The issue isn't our configuration - it's how Android decides to display keyboard names in the system UI.

I tested with "Akshay" → "akshay" confirms that Android automatically lowercases any custom string we provide.

If you have any suggestions on how to work around this, let me know! 

Screenshot:

![Image](https://github.com/user-attachments/assets/2553f9e2-4afc-4a95-ab15-d2dc2794d53b)

andrewtavis:
Thanks for the investigation and the screenshot, @Akshaykomar890! I think that it makes sense that we do the locale code as a lower case language name would be unexpected. Ideally we would have gotten it as the actual name of the language, but localized into the user's device language is a solid second option 😊

Please feel free to open a PR for this! :)

PR #500, opened 2025-10-01
Fix: replace custom language strings with locale codes (#493)

Akshaykomar890:
> **Fix:replace custom language strings with locale codes (#493)**

- Use standard locale codes (de, sv, it, fr) instead of custom language strings to fix lowercase display issue in Android keyboard selection.

> **Below are the screen shots for the ref:**

WhatsApp Image 2025-10-02 at 12 10 59 AM

### DIFF
--- a/app/src/main/res/xml/method_french.xml
+++ b/app/src/main/res/xml/method_french.xml
@@ -3,6 +3,6 @@
     android:icon="@mipmap/ic_launcher" >
     <subtype
         android:label="1. Français"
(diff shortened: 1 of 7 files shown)


andrewtavis (approved):
Checked and the changes are working wonderfully, @Akshaykomar890 😊 Thanks so much for a great first contribution!

CC @angrezichatterbox that the changes were reverted as we discussed and we're not good on this end :)

Issue #409, opened 2025-05-26
Add alphabetical autocompletion as words are being typed


andrewtavis:
### Terms

- [x] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Android/issues?q=is%3Aissue+label%3Afeature)
- [x] I agree to follow Scribe-Android's [Code of Conduct](https://github.com/scribe-org/Scribe-Android/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

An important functionality that's still missing from Scribe-Android is autocompletions that are displayed over the keyboard as the user types a word, but before they press space (this is covered by #408 that does autosuggestions). The word that the user is currently typing should be checked and if it's a key in the completions data, and if so then the next three words alphabetically in this dataset should be returned to the user in the autosuggestion/completion keys above the keyboard. Clicking one of these should insert it into the text proxy and run other commands like autosuggestions, noun annotations and emoji annotations. If there are emojis associated with the last word, then the third autocompletion should not be displayed and it should be replaced by the emoji suggestion buttons.

- Note: Blocked by #400 and #404 

### Contribution

Happy to support here and check a PR once it's been opened! 🚀

andrewtavis:
Thanks for your interest in helping out here, @catreedle! I'll also loop in @angrezichatterbox as he might have some feedback here. I think that autosuggestions are already done at this point, and the whole point is we want to have completions added such that the user can click to complete a word they're typing. More explicitly:

- The example above is how autosuggestions should work, yes :)
- Say the user types "Bu", then we'd want one of the keys to be "Buch" or the next alphabetical word
- We'll need to finalize some part of the app such that those times that autosuggestions or autocompletions shouldn't conflict with one another are figured out (we can do this later though)

Big thing is, we're trying to finalize it so that as the user types the next words alphabetically are suggested. We can start from here, and then once there's some progress we can finalize all the fine tuning that we need to do 😊

andrewtavis:
@catreedle, sorry for the wait on this. I've just added 30 strings to the `autocomplete_lexicon` table in each database. Basically what we have in there now is `qwerty`, `asdfg`, `zxcvb`, `abcde`, `aaaaa`, etc in alphabetical order :) This should be enough for an initial feature, and then we can bring over the full autocomplete lexicon a bit later?

Let us know how this sounds and if there are any questions! 😊

andrewtavis:
[...]

PR #501, opened 2025-10-06
Autocompletion

catreedle:
### Contributor checklist

-   [x] I have tested my code with the `./grad/gradlew lintKotlin detekt testlew lintKotlin detekt test` command as directed in the [testing section of the contributing guide](https://github.com/scribe-org/Scribe-Android/blob/main/CONTRIBUTING.md#testing)

---

### Description

Feature: Autocompletion

- Adds real-time word suggestions as the user types.
- Uses a Trie (prefix tree) to store all words in memory.
- Trie is loaded once when the keyboard starts and refreshed when the user switches languages.
- Every typed letter triggers a prefix search in the Trie for instant suggestions.

Code has been built and tested on an emulator.
Looking forward to any feedback or reviews. 😊

###

-   #409


DeleMike:
Thanks, @andrewtavis, for the tip. I have tried this but I still cannot get it to work.


https://github.com/user-attachments/assets/a70b08e2-6d1e-4e6b-ac0e-bcbabbe2df2f

Can you do a screen recording?

app/src/main/java/be/scri/helpers/data/Trie.kt
DeleMike, line 1: just love this! haha

DeleMike (approved):
Well done @catreedle! 

All looks good to me @andrewtavis 🔥

catreedle:
hi @DeleMike, @andrewtavis, @angrezichatterbox 

thanks for all the feedback and for looking into this 😊

@angrezichatterbox can you please elaborate on the conflict between autocompletion and autosuggestion?
maybe I can look into it as well. thank you! 😊

catreedle:
> > @angrezichatterbox can you please elaborate on the conflict between autocompletion and autosuggestion?
> > maybe I can look into it as well. thank you! 😊
> 
> Right now when you start typing and there is a suggestion from auto completion and then you start typing and there is a quick change between the state of having an auto completion suggestion and not having any suggestion you get a glitch of quickly changing. Instead we could have priority based system being implemented for it. Hope it's clear now.

Oh, I see. For a split second while typing, it’s showing the BASE_AUTOSUGGESTIONS. When there’s no autocompletion suggestion, it falls back to showing BASE_SUGGESTIONS.
What should be shown in these two cases instead? Is that where the priority-based system comes in?

catreedle:
> yes @catreedle , thanks for this. I understand better. Then does this PR affect the autosuggestion? I wanna reproduce the error from my end.

For me, the glitch happens while typing, for example, going from a → aa. There’s a split second where the default autosuggestions (BASE_AUTOSUGGESTIONS) appear. The same thing happens when typing something with no autocomplete suggestions.

This is affecting the autocomplete behavior, but I’m not sure how it affects the autosuggestions part.

DeleMike:
https://github.com/user-attachments/assets/7ea75136-6dcc-4744-bc77-52e12df16c28

See my experience.

Issue #492, opened 2025-09-26
Don't color first alternate key option as Scribe blue


andrewtavis:
### Terms

- [x] I have searched [open and closed design issues](https://github.com/scribe-org/Scribe-Android/issues?q=is%3Aissue+label%3Adesign)
- [x] I agree to follow Scribe-Android's [Code of Conduct](https://github.com/scribe-org/Scribe-Android/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

Scribe-Android has two different ways to select alternate keys: The default where the user slides to select an option, and an option that's selectable from the menu where the alternate keys view persists and a key can be selected. When the user is dragging to select an alternate key, the key selected is made Scribe blue so that there's a visual indication of which character will be inserted. For the persistent view, as of now the first option is Scribe blue, which isn't needed as the user can select any of the keys.

This issue would remove the Scribe blue coloring of the first alternate key if the user has deselected the hold to select alternate keys default menu option for keyboards.
 
- Note: The PR that most recently worked on the needed files for this and set up the hold/non-hold alternate keys options is #485

### Contribution

Happy to support as needed and review a PR when one is opened! 😊

andrewtavis:
Sure thing, @Akshaykomar890! Looking forward to the contribution! 😊

PR #502, opened 2025-10-06
Fix: alternate key blue highlighting based on hold setting (#492)

Akshaykomar890:
- Only show blue color on first alternate key when "hold for alt keys" setting is enabled. 
- In persistent mode, all keys now show normal color.


andrewtavis (approved):
All's working really well here, @Akshaykomar890! Great to have you taking on more issues 😊

One thing, could we ask you to till out the pull request template that's autogenerated when you open a PR the next time? Really helps with the review process if we can get the issue linked and have PRs looking generally the same as far as where the information we need is :)

Issue #491, opened 2025-09-26
Add shadow to alternate keys view


andrewtavis:
### Terms

- [x] I have searched [open and closed design issues](https://github.com/scribe-org/Scribe-Android/issues?q=is%3Aissue+label%3Adesign)
- [x] I agree to follow Scribe-Android's [Code of Conduct](https://github.com/scribe-org/Scribe-Android/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

When a user long holds certain keys on Scribe keyboards there is a view that shows alternate keys like those that have accents. As of now there's no visual distinction that this view is above the keyboard. It would be best if this view had a grey shadow with a color (`3F3F46`) so that it works for light and dark mode.

- Note: The PR that most recently worked on the needed files for this and set up the hold/non-hold alternate keys options is #485

### Contribution

Happy to support the work here and review a PR! 😊

Femimat:
Can I work on this issue @andrewtavis ?

PR #503, opened 2025-10-07
Update background colors for keyboard popup (day/night themes)

Femimat:
### Contributor checklist

---

### Description

This pull request updates the background colors for the mini keyboard popup to match the app’s day and night themes.
Previously, the popup background color only adapted correctly in dark mode, but not in light mode.
This fix ensures consistent theming across both modes by updating the corresponding color values in colors.xml and colors-night.xml.

### Related issue

-   #491

### DIFF
--- a/app/src/main/res/values-night/colors.xml
+++ b/app/src/main/res/values-night/colors.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
-    <color name="you_keyboard_background_color">#191C1E</color>
+    <color name="you_keyboard_background_color">#3F3F46</color>
     <color name="you_keyboard_toolbar_color">@color/you_background_color</color>
     <color name="app_text_color">@color/dark_text_color</color>
     <color name="nav_bar_color">#000</color>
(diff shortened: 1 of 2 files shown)


andrewtavis:
Hey @Femimat 👋 I'm realizing that I might have been confused by the plans for this. Would we be able to add a small shadow to the views that's similar to what's under the keys? The background color for the alternate keys view should be the same as they keyboard background color. It'd be good if we had a shadow under it so it's clear that it's above they keyboard interface :)

Femimat:
Got it! Thanks for clarifying — I’ll update the popup view to match the keyboard background color and add a subtle shadow similar to the one under the keys.

andrewtavis:
Sounds good, @Femimat! Looking forward to the update 😊

angrezichatterbox:
> @angrezichatterbox: By alternate keys view I mean the view that pops up when you hold the `a`, `o`, `u`, etc keys that have accent characters. If we could change the color of that view to be the same as the special keys then this should be able to be closed. Any non-necessary changes to that could also be removed from this, but then I'm not sure if some of the current changes will be used other places 😊
> 
> Thanks for checking in on this! :)

It been long since I have worked on it so I got a little confused. I will get the fix ready in a while.

angrezichatterbox:
Screenshot 2025-11-22 at 2 16 58 PM
Screenshot 2025-11-22 at 2 13 38 PM


Would this be fine ?

angrezichatterbox:
The other colors blend in pretty well with the background of the keyboard.

Issue #499, opened 2025-09-29
Allow for capitalized inputs for commnds to return valid outputs


andrewtavis:
### Terms

- [x] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Android/issues?q=is%3Aissue+label%3Afeature)
- [x] I agree to follow Scribe-Android's [Code of Conduct](https://github.com/scribe-org/Scribe-Android/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

Something that should be finalized before the first release is that we currently are not allowing the user to pass capitalized values to all commands. An example of this is:

- In the German keyboard the user selects the Scribe key at the top left and then translate
- They type the word "Book" (note that "book" is in the database)
- The result is that they're told the word isn't in Wikidata
- Ideally we'd use the same method from Scribe-iOS
    - The capitalization of the input word is derived before the database is checked and saved in a variable
    - The input word is converted to lower case (except for German nouns)
    - This word is checked against the database
    - If the original word was capitalized, then the capitalization of the return value is returned
    - If the original word wasn't capitalized, then the normal return value is returned

### Contribution

Happy to support the planing for implementation here and review a PR once one is opened! 😊

anileshwar12:
Hey @andrewtavis 
I’d like to work on this issue, could you please assign it to me?
Just to confirm, should the behavior match the iOS implementation, including handling cases like German nouns?

Akshaykomar890:
Hey @andrewtavis ,

Just checking in on this issue. If @anileshwar12 is unable to continue, I'm still interested and ready to start working on it whenever you're ready to reassign.

Thanks!

andrewtavis:
Quick check in @Akshaykomar890, do you need any assistance in working on this issue? Please let us know if you need any support!

PR #512, opened 2025-11-25
Allow capitalized input

catreedle:
### Contributor checklist

---

### Related issue

This PR adds changes to Translate command and Plural command to normalize capitalized inputs to be recognized in the database, so it can return valid outputs. This also adds specific handling for German language, especially for its capitalized nouns.

-   #499

### DIFF
--- a/app/src/main/java/be/scri/helpers/data/TranslationDataManager.kt
+++ b/app/src/main/java/be/scri/helpers/data/TranslationDataManager.kt
@@ -54,15 +54,15 @@ class TranslationDataManager(
 
         val variants = mutableListOf<String>()
 
-        // Try exact match
+        // Try exact match of input.
         variants.add(word)
 
-        // Lowercase variant:
+        // Add lowercase variants.
         if (isGerman || isWordCapitalized(word)) {
             variants.add(word.lowercase())
         }
 
-        // German-only: canonical noun capitalization ("buch" → "Buch")
+        // Note: In German canonical noun is capitalization ("buch" → "Buch").
         if (isGerman) {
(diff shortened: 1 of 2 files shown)


catreedle:
thank you! looking forward to the feedback on this 😊

catreedle:
Hi @andrewtavis @DeleMike, so sorry for missing this. looking into this soon 😊

app/src/main/java/be/scri/helpers/data/TranslationDataManager.kt
```diff
@@ -47,9 +48,50 @@ class TranslationDataManager(
...
+                    queryForTranslation(db, sourceTable, destCode, lowerCaseWord)
+                } ?: ""
+
+            if (translatedLowerCaseWord.isNotEmpty()) {
+                return translatedLowerCaseWord.replaceFirstChar { it.uppercase() }
+            }
+        }
+
+        return ""
     }
```
angrezichatterbox: Also we need not open the db twice. In our case it doesn't matter much as our keyboard are handled separately but it is not a good practice. Something to keep in mind.

andrewtavis:
I'd agree as well that it would be good to start testing the base behavior of the app. The plan was to focus on this post release, but I think another epic or two that focus on tests for specific parts would be a good idea. Maybe @catreedle can write tests for what she's adding, and others can focus on testing current functionality 🤔

andrewtavis (approved):
praise: All's working very well here, @catreedle 😊 Well done! 🙌 Really appreciate the quick work and the testing of the solution 🚀

note: I think we need to look into the conjugations being returned also being capitalized if the entered text is capitalized. I'll make an issue for this and ping you all :)

Issue #515, opened 2025-12-09
App language resets to English when Indonesian is selected


catreedle:
### Terms

- [x] I have searched all [open bug reports](https://github.com/scribe-org/Scribe-Android/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [x] I agree to follow Scribe-Android's [Code of Conduct](https://github.com/scribe-org/Scribe-Android/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

When selecting Indonesian as the app language from App Settings, the app language is set to English instead.
Other languages apply correctly.

Steps to reproduce:
1. Open the app
2. Go to Settings → App language
3. Select Indonesian

Result: App language switches to English
Expected: App language switches to Indonesian

### Device type

Emulator Pixel API 33

### Versions

Android 13

PR #526, opened 2025-12-24
add values-in/string.xml

catreedle:
### Contributor checklist

---

### Description

Added values-in folder to support legacy Indonesian language code. Android uses both "in" (legacy) and "id" (modern) codes for Indonesian, but translations were only in values-id.

### Related issue

-   #515


andrewtavis:
question: @catreedle, how do we think we can achieve this in a more sustainable way? I think that this change should almost certainly go into [Scribe-i18n](https://github.com/scribe-org/Scribe-i18n). Is what you're saying that the contents should be the same for each file? Maybe we can localize the `id` file and then copy it over when we run the Python scripts? If `in` should be different than `id`, then how about we add a new language into Scribe-i18n? :)

andrewtavis:
Mapping out the work left for this issue here, @catreedle:
- Finalize the four scripts in Scribe-i18n such that `in` and `id` are synced by them
    - Maybe best via a dictionary that as of now just maps `id` to `in` such that we can add more key value pairs later
    - If someone localizes `in` on Weblate, then we can copy it to `id`
- Move your change in this PR over to Scribe-i18n
- Do the Android conversion script to generate the JSON for `in`
- Update Scribe-i18n and we'll close https://github.com/scribe-org/Scribe-i18n/issues/99
- Update Scribe-i18n within this branch and then we'll bring in the latest version to close this PR 😊

PR #595, opened 2026-03-31
Speed up unit tests

Femimat:
### Contributor checklist

- [x] I have tested my code with the `./gradlew lintKotlin detekt test` command as directed in the [testing section of the contributing guide](https://github.com/scribe-org/Scribe-Android/blob/main/CONTRIBUTING.md#testing)

---

### Description

1. Matrix Strategy: I added a matrix strategy to parallelize the tests for the Conjugate and Keyboards flavors. This allows both sets of tests to run at the same time on different runners.
2. Pre-building APKs: I added a Pre-build Android Test APKs step that runs before the emulator starts. This prevents the build process from competing for resources with the emulator, which can lead to timeouts and slower test execution.
3. Removed redundant delays: I removed the sleep 30 and sleep 10 commands, as android-emulator-runner handles waiting for the device. I also removed redundant animation settings that are already handled by the disable-animations: true flag.
4. Workflow Cleanup:

- Renamed the "Run Ktlint" step in the test job to "Run Unit Tests" to accurately reflect that it runs ./gradlew test.
- Adjusted timeout-minutes to a more appropriate 45 minutes per parallel job.

- #580


DeleMike (commented):
LGTM too. Can you share with me the time it takes now?

Earlier, the instrumentation tests usually took between `12 to 15` minutes

Femimat:
Hi @DeleMike, The tests now run in parallel: Conjugate took 7m and Keyboards took 13m. This brings the total wall-clock time down to 13 minutes, while providing much faster feedback for the first flavor.

[ci_android / build app (pull_request)](https://github.com/scribe-org/Scribe-Android/actions/runs/24707598855/job/72264278886?pr=595)
ci_android / build app (pull_request)Successful in 4m
[ci_android / collect_coverage (pull_request)](https://github.com/scribe-org/Scribe-Android/actions/runs/24707598855/job/72264278813?pr=595)
ci_android / collect_coverage (pull_request)Successful in 5m
[ci_android / Run Android Instrumentation Tests (Conjugate) (pull_request)](https://github.com/scribe-org/Scribe-Android/actions/runs/24707598855/job/72264278842?pr=595)
ci_android / Run Android Instrumentation Tests (Conjugate) (pull_request)Successful in 7m
[ci_android / Run Android Instrumentation Tests (Keyboards) (pull_request)](https://github.com/scribe-org/Scribe-Android/actions/runs/24707598855/job/72264278839?pr=595)
ci_android / Run Android Instrumentation Tests (Keyboards) (pull_request)Successful in 13m
[ci_android / Run Ktlint and Detekt (pull_request)](https://github.com/scribe-org/Scribe-Android/actions/runs/24707598855/job/72264278830?pr=595)
ci_android / Run Ktlint and Detekt (pull_request)Successful in 1m
[ci_android / Run Tests (pull_request)](https://github.com/scribe-org/Scribe-Android/actions/runs/24707598855/job/72264278854?pr=595)
ci_android / Run Tests (pull_request)Successful in 5m

DeleMike (approved):
Thank you so much @Femimat! 

I understand there may be minor factors affecting the speed of these tests. However, I feel like an average of 10-12 minutes is fair. 

Actually, when you physically do an instrumentation test, they take around the same time, if not longer.

Unless someone has other ideas, everything looks good here.
@angrezichatterbox, what do you think?

cc: @andrewtavis

.github/workflows/ci_android.yaml
```diff
@@ -91,12 +86,16 @@ jobs:
         if: success()
         with:
           name: latest-apk
-          path: "app/build/outputs/apk/core/debug/app-core-debug.apk"
```
andrewtavis, line 94: Sorry for the wait here, @Femimat :)

question: One thing that I'm wondering about is this line. We're changing this to reference conjugate, but then I'd expect either both keyboards and conjugate to be referenced or for us to not do a change. Could you let us know why we're doing the change from `core` to `conjugate` here?

CC @DeleMike in case this is more clear to you 😊

Femimat:
Hi, @andrewtavis. You're right! I updated it to conjugate because the core flavor was replaced by keyboards and conjugate in the current build configuration but if it would make sense i can use a wildcard: app/build/outputs/apk/*/debug/*.apk. This way, both variants will be available as artifacts.

andrewtavis:
Nice ok :) Do you want to send along that change, @Femimat?

Femimat:
Hi @andrewtavis, I’ve updated the PR. Please take another look when you have time. Thanks!

andrewtavis (approved):
praise: Looks good, @Femimat! Thanks for your reaction to the review :) Looking forward to the next contribution!

PR #604, opened 2026-04-26
feat: enforce changelog entry via CI check

Mounil2005:
### Contributor checklist

- [ ] I have tested my code with the `./gradlew lintKotlin detekt test` command as directed in the [testing section of the contributing guide](https://github.com/scribe-org/Scribe-Android/blob/main/CONTRIBUTING.md#testing)

---

### Description

Adds a new GitHub Actions workflow `ci_changelog_check.yaml` that enforces changelog discipline on every PR targeting `main`.

**What changed:**
- `.github/workflows/ci_changelog_check.yaml` -> new workflow file

**How it works:**
- Uses `tj-actions/changed-files` to detect whether `CHANGELOG.md` was modified in the PR
- If it was not updated, the workflow fails with a clear error message pointing contributors to update the changelog before merging
- Supports a `no-changelog` label that maintainers can apply to skip the check for CI fixes, typo corrections, or dependency bumps
- Triggers on `opened`, `reopened`, `synchronize`, `labeled`, and `unlabeled` so adding/removing the label immediately re-evaluates the check

**Testing:**
This is a pure YAML workflow file, the `lintKotlin detekt test` command does not apply. The check can be verified by opening a test PR without a changelog update and confirming it fails.

### Related issue

No related issue, this is a self-initiated improvement to the CI pipeline.


andrewtavis:
@axif0, @DeleMike: What's your two's opinion on this? I think that it could be nice if we could eventually migrate away from the maintainer checklists that we have for projects. Each checklist that we have has one box for tests passing and having been written as well as a box that the changelog has been updated. We haven't been doing the changelog for Scribe-Android yet, and tend to write the changelog for Scribe-Data as PRs with major changes come in.

Would be great if the testing checkbox was automatic given tests in CI and coverage being maintained, especially for the files being changed (maybe via `tj-actions/changed-files@v44` as is used here, but the focus on overall coverage for now is fine). For the changelog, I think there could be a way to automate this a bit better, but I'm just not sure if this is exactly what we're looking for. Ideally we'd come up with something that all of the projects could use 😊

Mounil2005:
> @axif0, @DeleMike: What's your two's opinion on this? I think that it could be nice if we could eventually migrate away from the maintainer checklists that we have for projects. Each checklist that we have has one box for tests passing and having been written as well as a box that the changelog has been updated. We haven't been doing the changelog for Scribe-Android yet, and tend to write the changelog for Scribe-Data as PRs with major changes come in.
> 
> Would be great if the testing checkbox was automatic given tests in CI and coverage being maintained, especially for the files being changed (maybe via `tj-actions/changed-files@v44` as is used here, but the focus on overall coverage for now is fine). For the changelog, I think there could be a way to automate this a bit better, but I'm just not sure if this is exactly what we're looking for. Ideally we'd come up with something that all of the projects could use 😊

Thanks for the detailed feedback! A few thoughts on the points raised:

Cross-project reusability - happy to convert this into a reusable workflow using workflow_call so it can live in scribe-org/.github and be called from all Scribe projects with minimal config. That would make it a one-liner to opt in from any repo.

Hard-fail vs. soft warning - since Scribe-Android hasn't been enforcing changelog entries yet, a hard fail might be too abrupt for contributors. One middle ground: keep the CI failure but also have the workflow post a PR comment when it fails, explaining exactly what to add and where (e.g. "Please add an entry under the latest version heading in CHANGELOG.md"). That way contributors aren't just staring at a red check with no guidance.

Test checkbox automation - great idea as a follow-up. tj-actions/changed-files could be used to check whether changed .kt files have corresponding test coverage. Happy to explore that in a separate PR once the changelog piece is settled.

Would it make sense to first agree on the changelog enforcement approach here, then tackle reusability and test automation as follow-ups? Would also love to hear what @axif0 and @DeleMike think!

DeleMike (commented):
Hi @Mounil2005, this is a good PR; it will have its benefits. I do think that having a PR that checks for a compulsory changelog entry is good. 

As @andrewtavis noted, this is a nice-to-have and not yet a priority.

DeleMike:
How about looking into adding more test coverage for the app? is that something you can work on?

Mounil2005:
> How about looking into adding more test coverage for the app? is that something you can work on?

Thanks for the suggestion @DeleMike! I would surely love to work on that.
After looking into the codebase, the biggest gaps are:

**Core keyboard handlers:** AutocompletionHandler, BackspaceHandler, KeyHandler, SpaceKeyProcessor, and SuggestionHandler are completely untested despite being the heart of the app. These are pure logic so unit tests with the existing JUnit 5 + Mockk setup would work well.
**Data managers :** ConjugateDataManager, GenderDataManager, PluralFormsManager, Trie etc. all have zero coverage. Robolectric is already declared in the project so SQLite-backed ones can be tested without an emulator.
**Language-specific helpers:** 8 language implementations, none tested. Even basic smoke tests to verify each language loads and returns valid data would be a solid start.
**Conjugate flavor:** CI already runs connectedConjugateDebugAndroidTest but there are no actual test files for it, so that job is effectively a no-op.

which one should i move ahead with?

DeleMike:
Okay, kindly wait for @angrezichatterbox to get back to you on this. 

Or @andrewtavis, what are your thoughts on this?

Mounil2005:
> Hi @Mounil2005, this is a good PR; it will have its benefits. I do think that having a PR that checks for a compulsory changelog entry is good.
> 
> As @andrewtavis noted, this is a nice-to-have and not yet a priority.

Thanks @DeleMike! Completely understand it's not a top priority right now. Since the implementation is already in place though, it might be a nice opportunity to merge it as a quiet foundation, it won't change anything for existing contributors and can always be tightened up later when the team feels ready.
I was also thinking of extending it to post a friendly PR comment when the check fails, pointing contributors to exactly where and how to add a changelog entry. That way it feels more like a helpful nudge than a hard blocker.
Of course, totally open to whatever direction the team prefers!

DeleMike:
No problem @Mounil2005, let’s wait for @andrewtavis decision.

andrewtavis:
The one thing I'm wondering about here is whether we need a changelog update for each PR. But then I don't think we'd be able to have it know whether a change is large enough for a changelog entry 🤔

The final question I have here is whether we should look into changelog automation. I've seen things like this before, where the issue number from the PR and the commit message go into the changelog. I'm not sure if this would really be a benefit. @Mounil2005, could you take a look into some options for this and let us know what you think?

Once we check the changelog auto update process, if we don't choose to go that way we can bring this in :)

Mounil2005:
> The one thing I'm wondering about here is whether we need a changelog update for each PR. But then I don't think we'd be able to have it know whether a change is large enough for a changelog entry 🤔
> 
> 
> 
> The final question I have here is whether we should look into changelog automation. I've seen things like this before, where the issue number from the PR and the commit message go into the changelog. I'm not sure if this would really be a benefit. @Mounil2005, could you take a look into some options for this and let us know what you think?
> 
> 
> 
> Once we check the changelog auto update process, if we don't choose to go that way we can bring this in :)

Sure, I will look into it and revert back soon!

Mounil2005:
Hi @andrewtavis, @DeleMike

As per the feedback above, I researched options for automating changelog updates for Scribe. Here’s what I found:

1. Auto Changelog (e.g., mikepenz/release-changelog-builder-action):

Aggregates PR titles, labels, or commit messages to generate or update the CHANGELOG.md automatically.
Flexible: Works with natural commit messages/titles; customizable templates; low setup overhead.
Fits our current PR-based workflow, doesn’t require changing how we write commits.

2. Semantic Release:

Fully automates versioning, changelog, and even publishing, but requires strict [Conventional Commits](https://www.conventionalcommits.org/) across all contributors.
Enforces a next-level release process (auto-tagging, publishing).
Would require significant habit change for all contributors.

andrewtavis:
@Mounil2005: I've just finished writing the MVP changelogs for the applications, so I think we're good to start enforcing changelog entries on PRs. I've merged main into this branch and added your changelog entry in. The one thing is that we now have two changelogs for the project - one for the keyboards and one for the conjugate app in `CHANGELOG_CONJUGATE.md`. Would you be able to update this PR to also check for that file? I think we can just assert that one or the other file has been updated.

Aside from this, you'd be welcome to open this same PR in the other projects:
- Scribe-iOS: We also want it checking a conjugate changelog
- Scribe-Data: Just the changelog file
- Scribe-Server: Just the changelog file

I don't think we need this workflow on other repos at this time :)

CC @DeleMike and @axif0 for if either of you have other thoughts here!

andrewtavis:
And I'm realizing we didn't provide feedback on your investigation. Firstly, thank you for it! Really great to see the options. I think that what you have here is simple and works well for the way the community operates. Let's do the necessary update for a second changelog file for this and iOS, but aside from that I think your solution is great 😊

andrewtavis (approved):
praise: Now that we have the changelogs written, @Mounil2005, we're thinking that adding this in would be a very solid addition. Thanks so much for the idea! I'll go ahead and add this in to the other repos so we have CI parity 😊

Issue #551, opened 2026-03-03
Implement Autosuggestions and Autocompletions in the Keyboard


angrezichatterbox:
### Terms

- [x] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Android/issues?q=is%3Aissue+label%3Afeature)
- [x] I agree to follow Scribe-Android's [Code of Conduct](https://github.com/scribe-org/Scribe-Android/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

Refactor the current autosuggestion and autocompletion system to replace the existing database based approach with a hybrid Trie + LLM architecture similar to the AOSP keyboard. The Trie will handle efficient prefix-based candidate generation, while the LLM layer will provide contextual reranking and next-word prediction to improve relevance and accuracy. This change aims to enhance suggestion quality, contextual awareness, and scalability while maintaining low-latency performance suitable for mobile environments.

### Contribution

I would be working on this. Happy to hear suggestions and improvements to the current plan :)

andrewtavis:
@angrezichatterbox, are we good to close this? Feel free to close if we are :)

One final check, we're still reverting to alphabetical autocompletions via #409 if we don't have the files added in from #610, right? And the plan is we'll switch over autosuggestion backups and maybe the primary autosuggestions to trained language models in the future, correct?

andrewtavis:
@angrezichatterbox, quick ping on the above :) Feel free to close this if we're still reverting to the alphabetical autocomplete in the current implementation! 😊

angrezichatterbox:
Yeah we are falling back to that incase we dont get any data from the trie based selection. We could close this for now. 

> And the plan is we'll switch over autosuggestion backups and maybe the primary autosuggestions to trained language models in the future, correct?

Yeah that would be something to make this application be much better : )

Issue #521, opened 2025-12-20
Some keyboard functionality doesn't work on first load


andrewtavis:
### Terms

- [x] I have searched all [open bug reports](https://github.com/scribe-org/Scribe-Android/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [x] I agree to follow Scribe-Android's [Code of Conduct](https://github.com/scribe-org/Scribe-Android/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

An issue that we're dealing with for Scribe-Android is that we are unable to do certain actions on the keyboard on first load. Specifically, the delete key does not work until the user has navigated away from the keyboard being shown and then back to the keyboard. If memory serves me, there are also some issues with commands being processed on first keyboard load.

### Device type

Phones and tablets

### Versions

Current main branch

angrezichatterbox:
For me right now all the keys are working. Could you check wheather it is working for you after the refractor merge and all. 
@andrewtavis

andrewtavis:
Will do, @angrezichatterbox!

andrewtavis:
As noted in #610, @angrezichatterbox, we're once again experiencing a situation where noun annotations aren't being shown to the user. We can see them after the plural command, but not after typing a word. Do you want to look into that, and with the PR for that we can do a final check for all functionality working on first keyboard load?

PR #610, opened 2026-05-28
Implement Auto suggestions 

angrezichatterbox:
### Contributor checklist

- [x] I have tested my code with the `./gradlew lintKotlin detekt test` command as directed in the [testing section of the contributing guide](https://github.com/scribe-org/Scribe-Android/blob/main/CONTRIBUTING.md#testing)

---

### Description

This PR implements the autosuggestion in the Scribe-Android Keyboard. 

### Related issue

- #551


andrewtavis:
So amazing to have this open, @angrezichatterbox! 💙

andrewtavis:
I'll try to get to the review on Saturday :)

angrezichatterbox:
> I'll try to get to the review on Saturday :)

I have kept the Android file licenses as it is. That would be fine right.

andrewtavis:
I think that that would be appropriate. Thanks for flagging this! We definitely should have an expansive reference, and this seems like an appropriate way to do it :)

andrewtavis:
Looks like there's a minor conflict here now, @angrezichatterbox :) Can we resolve that and then I'll look to bring this in next? 😊

angrezichatterbox:
@andrewtavis  it is ready for review.

andrewtavis:
Thanks, @angrezichatterbox! I'll get to this sometime today :)

andrewtavis:
Autosuggestions are working really well, @angrezichatterbox :) One thing to note though is that this is now breaking noun annotations. This isn't an issue with the data, as when I execute a plural command I still get a `PL` annotation. Can you look into why the noun annotations aren't being displayed?

andrewtavis:
I'm realizing that the above is an issue on main as well... I'll look to bring this in and then we can handle this in #521 :)

andrewtavis (approved):
praise: This really is a massive undertaking, @angrezichatterbox :) Great work to get all of this set up!

note: We should discuss sometime this week how we want to make sure that these files are maintainable. Maybe we need to remove some of the files that we're not using as they're for languages that we don't have implemented right now? Let's discuss in the call later this week 😊

PR #648, opened 2026-07-06
feat: highlight autocomplete suggestion when it matches typed word

### LINKED ISSUE #407: Add noun gender annotations under autosuggestions and autocompletions
### Terms

- [x] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Android/issues?q=is%3Aissue+label%3Afeature)

### Description

A further feature of Scribe-iOS that Scribe-Android doesn't have is annotation of nouns in autosuggestions and autocompletions. An example of this in Scribe-iOS is:

Image

The colored rectangles below the words above tells the user that the first word is neutral and the second and third are feminine. An example of this can be seen in the [Figma designs](https://www.figma.com/design/c8945w2iyoPYVhsqW7vRn6/scribe_public_designs?node-id=4277-4434&t=TfEmwG7pqY7LkqiV-1). We want a blue line under masculine nouns, a red line under feminine nouns, a green line under neuter nouns, a purple line under common nouns and an orange line under plural nouns.

This issue would add these markers to Scribe-Android, where if a word is not a noun there would be no annotation, and if it is a noun then it would have the color of its noun gender displayed below it. If a noun has more than one gender - i.e. the word `Schild` in German that is both masculine and neutral, then it would display the rectangle as half blue and half green.

- Note: This issue is blocked by #400, #408, #409 and #551

### Contribution

Happy to support and open review a PR when one is opened! 🚀

### LINKED ISSUE #410: Highlight autocompletion if it is the word the user is typing
### Terms

- [x] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Android/issues?q=is%3Aissue+label%3Afeature)

### Description

Once work for #409 is complete, we'll want to add a small feature that's included in Scribe-iOS, which is if the user has typed the next autocompletion, then this word is highlighted. This can be seen in the image below where the user has typed `Buch`, which is the first completion:

Image

This indicates to the user that if they press the completion that nothing will happen (or we should run the insert, but this should just replace the word with itself).
 
- Note: This issue is blocked by #409

### Contribution

Happy to discuss this and review a PR when one is opened! 🚀

Mounil2005:
### Contributor checklist

- [x] This pull request is on a separate branch and not the main branch
- [x] I have tested my code with the `./gradlew lintKotlin detekt test` command as directed in the testing section of the contributing guide

---

### Description

Once the user has typed enough of a word that the first autocomplete suggestion exactly matches it (e.g. typing "Buch" when "Buch" is itself a suggestion), that suggestion chip is now highlighted (bold, accent blue text) to indicate that tapping it is effectively a no-op, matching the existing Scribe-iOS behavior described in the issue.

- `app/src/keyboards/java/be/scri/services/GeneralKeyboardIME.kt`: `setSuggestionButton` and `setAutocompleteButton` now accept the word currently being typed. If a suggestion's text exactly matches it (case-insensitive), the button is rendered bold using `R.color.theme_scribe_blue` instead of the default text color. Tap behavior is unchanged — it still deletes the current word and re-inserts the suggestion (a no-op when they're equal).
- `app/src/keyboards/java/be/scri/helpers/AutocompletionHandler.kt`: passes the current word through to `updateAutocompleteSuggestions` so the exact-match check has what it needs.

Tested by running `./gradlew lintKotlin detekt test` (all green) and `./gradlew :app:compileKeyboardsDebugKotlin`.

### Related issue

- Closes #410
- Closes #407

### DIFF
--- a/app/src/keyboards/java/be/scri/helpers/AutocompletionHandler.kt
+++ b/app/src/keyboards/java/be/scri/helpers/AutocompletionHandler.kt
@@ -49,7 +49,7 @@ class AutocompletionHandler(
                 val completions = ime.getAutocompletions(currentWord, limit = 5)
 
                 if (completions.isNotEmpty()) {
-                    ime.updateAutocompleteSuggestions(completions)
+                    ime.updateAutocompleteSuggestions(completions, currentWord)
                 } else {
                     ime.clearAutocomplete()
                 }
--- a/app/src/keyboards/java/be/scri/helpers/AutocompletionHandler.kt
+++ b/app/src/keyboards/java/be/scri/helpers/AutocompletionHandler.kt
@@ -46,10 +46,10 @@ class AutocompletionHandler(
                     return@Runnable
                 }
 
-                val completions = ime.getAutocompletions(currentWord, limit = 5)
+                val result = ime.getAutocompletions(currentWord, limit = 5)
 
-                if (completions.isNotEmpty()) {
-                    ime.updateAutocompleteSuggestions(completions, currentWord)
+                if (result.completions.isNotEmpty()) {
+                    ime.updateAutocompleteSuggestions(result.completions, result.highlightedSuggestion)
                 } else {
                     ime.clearAutocomplete()
                 }
--- a/app/src/keyboards/java/be/scri/helpers/SpaceKeyProcessor.kt
+++ b/app/src/keyboards/java/be/scri/helpers/SpaceKeyProcessor.kt
@@ -61,6 +61,13 @@ class SpaceKeyProcessor(
      * @param wasLastKeySpace true if the previous key pressed was a space.
      */
     private fun handleSpaceOutsideCommandBar(wasLastKeySpace: Boolean) {
+        if (ime.tryInsertHighlightedAutocompleteSuggestion()) {
+            val wordBeforeSpace = ime.getLastWordBeforeCursor()
+            suggestionHandler.processLinguisticSuggestions(wordBeforeSpace)
+            suggestionHandler.processWordSuggestions(wordBeforeSpace)
+            return
+        }
+
         val periodOnDoubleTapEnabled = PreferencesHelper.getEnablePeriodOnSpaceBarDoubleTap(context = ime, ime.language)
         val ic = ime.currentInputConnection ?: return
         val wordBeforeSpace = ime.getLastWordBeforeCursor()
--- a/app/src/keyboards/java/be/scri/services/GeneralKeyboardIME.kt
+++ b/app/src/keyboards/java/be/scri/services/GeneralKeyboardIME.kt
@@ -1885,8 +1885,11 @@ abstract class GeneralKeyboardIME(
 
         highlightedAutocompleteSuggestion = highlightedSuggestion
 
-        setAutocompleteButton(uiManager.binding.conjugateBtn, completion1, highlightedSuggestion)
-        setAutocompleteButton(uiManager.binding.translateBtn, completion2, highlightedSuggestion)
+        // Visual left-to-right order of these buttons is translateBtn, conjugateBtn, pluralBtn
+        // (see input_method_view.xml's constraint chain), so completion1 -- the one that may be
+        // highlighted -- must go in translateBtn to actually land in the first, leftmost slot.
+        setAutocompleteButton(uiManager.binding.translateBtn, completion1, highlightedSuggestion)
+        setAutocompleteButton(uiManager.binding.conjugateBtn, completion2, highlightedSuggestion)
         setAutocompleteButton(uiManager.pluralBtn!!, completion3, highlightedSuggestion)
 
         uiManager.binding.separator1.visibility = View.VISIBLE
--- a/app/src/keyboards/java/be/scri/services/GeneralKeyboardIME.kt
+++ b/app/src/keyboards/java/be/scri/services/GeneralKeyboardIME.kt
@@ -159,6 +159,12 @@ abstract class GeneralKeyboardIME(
     private var currentEnterKeyType: Int? = null
     private var isNumericKeyboardActive: Boolean = false
 
+    /**
+     * The autocomplete suggestion currently shown highlighted (if any), so a following space
+     * press can insert it in place of what the user has typed, matching Scribe-iOS behavior.
+     */
+    private var highlightedAutocompleteSuggestion: String? = null
+
     internal var currentState: ScribeState = ScribeState.IDLE
     internal var invalidCommandSource: ScribeState = ScribeState.IDLE
 
@@ -1846,10 +1852,12 @@ abstract class GeneralKeyboardIME(
         currentWord: String? = null,
     ) {
         if (currentState != ScribeState.IDLE) {
+            highlightedAutocompleteSuggestion = null
             uiManager.disableAutoSuggest(language)
             return
         }
         if (completions.isNullOrEmpty()) {
+            highlightedAutocompleteSuggestion = null
             uiManager.disableAutoSuggest(language)
             return
         }
@@ -1857,6 +1865,9 @@ abstract class GeneralKeyboardIME(
         val completion1 = completions.getOrNull(0) ?: ""
         val completion2 = completions.getOrNull(1) ?: ""
         val completion3 = completions.getOrNull(2) ?: ""
+
+        highlightedAutocompleteSuggestion =
+            completion1.takeIf { it.isNotBlank() && it.equals(currentWord, ignoreCase = true) }
 
         setAutocompleteButton(uiManager.binding.conjugateBtn, completion1, currentWord)
         setAutocompleteButton(uiManager.binding.translateBtn, completion2, currentWord)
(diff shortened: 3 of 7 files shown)


andrewtavis:
praise: Thanks so much for the PR, @Mounil2005! We'll try to get to the review in the coming days :) 

note: For UI changes like this, would you be able to send along a screenshot that includes the change? If it's something like this, then maybe there's no need for a screenshot of the state before the change, but before and after screenshots are also quite helpful as a reviewer in knowing that the work has been completed before going into the code. Just suggesting! 😊

Mounil2005:
While testing this locally with Scribe German, I found that the native dictionary engine 
(NativeSuggestionEngine.getAutocompletions) explicitly filters out the exact typed word from its completions, it's designed to only suggest extensions of what you've typed (e.g. typing Buch only returned buchen, Buchhändler, Bücherei, never Buch itself). That meant the highlight-on-exact-match feature could never actually trigger through the primary suggestion path, since the exact match never reached the UI.

Rather than removing that filter (which would make "exact match shows first" dependent on the native dictionary's internal ranking, not guaranteed), I used the dictionary's existing isValidWord/isInDictionary check separately: if what the user has typed is itself a real word, it's now always shown as the first suggestion, with the remaining slots filled by real completions, deterministically matching the behavior in the issue's screenshot rather than hoping the ranking happens to put it first.

Tested on-device (emulator) with Scribe German across several words (Buch, Haus, Auto, Zeit, and various prefixes)...the exact match consistently shows first and highlighted (bold, accent blue) whenever it's a valid word, while non-exact prefixes correctly show only extensions with no highlight.

Screenshot 2026-07-07 032903
Screenshot 2026-07-07 032934
Screenshot 2026-07-07 033007

andrewtavis:
note @Mounil2005, @@Priyanshikaa1111: I'm thinking that it might be for the best if we move the word that the user is typing to the first autocompletion sole and just change the background to be highlighted rather than the text being blue. This at least is a common behavior in iOS keyboards. Specifically what this allows us to do is also have a space functionality where when the word is highlighted, pressing space will insert it.

The way that this works in Scribe-iOS is the following:
- Say that we have an English words registry with the words "house", "houses", "drum" and "dolphin"
- The user types "hous" and "house" is the first autocomplete, but not highlighted
- The user types "e" and "house" is the first autocomplete (not really an autocomplete anymore) and is highlighted
- There's really no difference to pressing space here
- If the user types "do", then "dolphin" is the only word that can follow
- The word "dolphin" is now the first autocomplete
- If the user presses space, then we'd insert "dolphin" in this case as it was highlighted
- We'd also make "do" the first autocomplete for if the user actually did want to type the word that they'd typed and just wanted a space instead of an autocomplete

The example above of "do" -> "dolphin" would never happen, so this example is mainly for larger words where once a user types enough of it, then it's obvious that they're typing it and we might as well insert the rest on space for them. The indication that this is going to happen is the autosuggestion being highlighted :)

How does this sound to the both of you?

CC @angrezichatterbox for this discussion 👋

Mounil2005:
> note @Mounil2005, @@Priyanshikaa1111: I'm thinking that it might be for the best if we move the word that the user is typing to the first autocompletion sole and just change the background to be highlighted rather than the text being blue. This at least is a common behavior in iOS keyboards. Specifically what this allows us to do is also have a space functionality where when the word is highlighted, pressing space will insert it.
> 
> The way that this works in Scribe-iOS is the following:
> 
> * Say that we have an English words registry with the words "house", "houses", "drum" and "dolphin"
> * The user types "hous" and "house" is the first autocomplete, but not highlighted
> * The user types "e" and "house" is the first autocomplete (not really an autocomplete anymore) and is highlighted
> * There's really no difference to pressing space here
> * If the user types "do", then "dolphin" is the only word that can follow
> * The word "dolphin" is now the first autocomplete
> * If the user presses space, then we'd insert "dolphin" in this case as it was highlighted
> * We'd also make "do" the first autocomplete for if the user actually did want to type the word that they'd typed and just wanted a space instead of an autocomplete
> 
> The example above of "do" -> "dolphin" would never happen, so this example is mainly for larger words where once a user types enough of it, then it's obvious that they're typing it and we might as well insert the rest on space for them. The indication that this is going to happen is the autosuggestion being highlighted :)
> 
> How does this sound to the both of you?
> 
> CC @angrezichatterbox for this discussion 👋

That sounds great, thanks for the detailed writeup! I've implemented all three pieces:

1. **Background highlight instead of blue text:** the highlighted suggestion now shows a rounded, tinted background box behind the text, matching the Scribe-iOS reference image, instead of the bold blue text from the first version of this PR.
2. **Highlighted suggestion always in the first slot + space inserts it:** pressing space now checks if a suggestion is highlighted and, if so, inserts it (with a trailing space) in place of the word being typed, instead of just committing a literal space.
3. **Broader trigger for "obvious" completions:** highlighting no longer only fires when the user has typed a word out in full. If only one dictionary word can possibly complete what's been typed so far (the do → dolphin case), that word is now promoted to the first slot and highlighted, while what the user actually typed (do) is still offered as another suggestion in case they wanted to keep typing their own word.

I'll be attaching images shortly to show this working end-to-end.

Mounil2005:
image

While testing the highlighting behavior across different word lengths, I found something worth flagging separately: the native suggestion engine's completions become increasingly unrelated to the actual typed prefix as the word gets longer, across all languages (not German-specific).

Example: typing dolp in English returns mol, DSL, Sol, AOL....none of which start with "dolp" and Dolphin never shows up as a candidate at all, despite being the obvious match. Shorter prefixes (dol) still look reasonable; the divergence grows with length.

Root cause: [ComposedData.createForWord()](vscode-webview://0bgc0udb20cffbnu8h3peips42jqutsuaal8t45n86oj171m18kl/blob/main/app/src/main/java/be/scri/latin/common/ComposedData.kt#L49-L61) generates random touch coordinates (Random.nextBits(2)) for every typed character instead of real ones, and this gets paired with a completely empty ProximityInfo (keyCount=0, no real key geometry). The underlying native engine is a spatial/gesture-typing correction algorithm, it ranks candidates partly by touch-point proximity, so feeding it random noise corrupts the ranking, with more accumulated noise (longer words) causing more drift from the actual prefix.

This predates this PR....it just wasn't visible before since the raw completions were only ever used for background suggestion chips. Step 3 of this PR (highlighting the single unambiguous candidate) depends on completion precision, which is what surfaced it.

Fixing it properly likely needs either real per-key coordinates for each keyboard layout or a suggest-mode that bypasses spatial correction for pure prefix lookup,both feel out of scope for #410, so flagging it here for a separate issue rather than trying to fix it in this PR. Happy to file it as its own issue if that's preferred.

Mounil2005:
As a mitigation for the above (while the deeper fix is still open), I added a startsWith(prefix, ignoreCase = true) filter to NativeSuggestionEngine.getAutocompletions(), so any suggestion shown or highlighted is now guaranteed to actually extend what the user typed. This specifically matters for the "single unambiguous candidate" highlight, which would otherwise have trusted and potentially auto-inserted (via space) an off-prefix candidate.

Verified on-device: dolphi now correctly returns dolphin's, dolphin, dolphins instead of the earlier junk.

Mounil2005:
image
image

some images after the final implementation.

andrewtavis:
Hey @Mounil2005 and @Priyanshikaa1111 👋 Looks like we're also doing the work for #407 here :) A few points:

- I would have the whole highlight have rounded corners and have a bit more height while also having the gender annotation line having a bit less width that the highlight
    - This is how we're doing it in iOS and it works quite well
    - See iOS screenshot below from the issues

Image

- Looks like we're assigning annotations for all words and that red (feminine) is the default?
    - Many of these words shouldn't have an annotation, and many of them are not feminine words
    - The annotations should be based only on the data in the nouns table
    - We're still in the process of finalizing the data process, so there might be some missing data now
    - Do your best with the data that's there and find a language that does have nouns and genders for testing, if German doesn't have them right now :)

CC @angrezichatterbox for the conversation here 😊

angrezichatterbox:
@andrewtavis We wouldn't have a lot of suggestions right that follow this pattern as we are using autosuggestions in the android keyboard.

angrezichatterbox:
So for every word we type we would be checking the autosuggestion data and doing a query to see the word gender right. We would have to make sure that the keyboard doesn't have to do a lot of db operations as well.

angrezichatterbox:
> This predates this PR....it just wasn't visible before since the raw completions were only ever used for background suggestion chips. Step 3 of this PR (highlighting the single unambiguous candidate) depends on completion precision, which is what surfaced it.
> 
> Fixing it properly likely needs either real per-key coordinates for each keyboard layout or a suggest-mode that bypasses spatial correction for pure prefix lookup,both feel out of scope for #410, so flagging it here for a separate issue rather than trying to fix it in this PR. Happy to file it as its own issue if that's preferred.

I think there is already a pure prefix based check in the codebase. We would have to make slight changes to the exising autosuggestions to be more deterministic.

andrewtavis:
> I think there is already a pure prefix based check in the codebase. We would have to make slight changes to the exising autosuggestions to be more deterministic.

Yes, this is true. We should be using this instead of something new.

> So for every word we type we would be checking the autosuggestion data and doing a query to see the word gender right. We would have to make sure that the keyboard doesn't have to do a lot of db operations as well.

Very true! This needs to be optimized.

> @andrewtavis We wouldn't have a lot of suggestions right that follow this pattern as we are using autosuggestions in the android keyboard.

This highlight is specifically for autocompletions. Is it possible to have the autocompletions be a bit more alphabetical and then use the autosuggestions from the current implementation?

Mounil2005:
Thanks for sharing the #407 spec details, that clarified exactly what was needed, so we went ahead and implemented it here rather than leaving it as a follow-up.

What we misunderstood initially: we'd used the annotateGreen/annotateRed color values to mean "this is the highlighted/obvious autocompletion" vs. "these are alternatives", not realizing those colors are already the established gender-annotation scheme. That's fixed now.

What's implemented:
- Underline color is now driven by the suggestion word's actual grammatical gender, looked up via the same nounKeywords map the app already loads once per session, so no extra DB queries per keystroke (addressing the performance concern above). Reuses the existing AnnotationTextUtils.handleColorAndTextForNounType mapping: masculine=blue, feminine=red, neuter=green, common=purple, plural=orange.
- Words with no known gender (verbs, non-nouns, or anything not yet in the nouns table) get no underline at all, no more defaulting to red/feminine for everything.
- Nouns with more than one gender (e.g. German "Schild" = masculine + neuter) get a split underline, half one color, half the other, matching the #407 description exactly.
- The "highlighted/obvious completion" background box from #410 is completely independent of this, a suggestion can be highlighted, gendered, both, or neither.
On the deterministic ordering / alphabetical / rounded corners feedback, all also implemented: the SQL/Trie lexicon is now tried before the native dictionary engine, results are genuinely alphabetical (fixed a bug where the Trie wasn't actually sorted before), and the highlight box now has more pronounced rounded corners with a narrower underline beneath it, matching the iOS reference more closely.

One caveat: since the nouns/lexicon table data is still being finalized per your note, gender annotations and completions may have gaps for some words/languages until that data fills in, that's expected on our end, not a bug.

Mounil2005:
image
image
some images

andrewtavis:
This is looking much better now, @Mounil2005 and @Priyanshikaa1111! Thank you both for the efforts here :) We'll look to bring this in soon 😊

CC @angrezichatterbox - lots of progress here. Would be great if you can do an initial check here :)

Mounil2005:
Hey @andrewtavis @angrezichatterbox , is there any update on this?

andrewtavis:
Thanks for checking in here, @Mounil2005! We're working on a few other PRs now, but I think once #660 is in we can move to this one 😊

Mounil2005:
> Thanks for checking in here, @Mounil2005! We're working on a few other PRs now, but I think once #660 is in we can move to this one 😊

oh got it! Thanks!

Mounil2005:
Hello, is there any update on this?

andrewtavis:
Hey @Mounil2005 👋 We have two PRs ahead of this one and then it'll be the next to bring in. Sorry for your wait. We just want to make sure that autosuggestions are solid before we bring this in. For now, if you rebase on main that'd be very helpful!

